Issue #29, opened 2026-03-07
Add a configurable push-to-record delay to differentiate between pressing and holding


jeffreyrr:
I would like a 1-2 second audio recording delay in order to differentiate between pressing and holding the recording activation button.

PR #30, opened 2026-03-07
Add hotkey hold-to-record delay setting

jeffreyrr:
Introduce a configurable hold delay for the push-to-talk hotkey. Adds a new AppState property (hotkeyHoldDelaySeconds) persisted to UserDefaults (key: hotkey_hold_delay_seconds) with values clamped to 0–5 and a default of 1s. Implements a DispatchWorkItem (holdToRecordWorkItem) to start recording only after the configured delay on hotkey down and cancel if released early. Exposes a "Hold delay" settings card with a Picker (0–5 seconds) in SettingsView to allow users to change the delay.

PR to address #29


marcbodea:
@zachlatta I think this should be closed as #34 adds the delay

PR #34, opened 2026-03-13
Add configurable dictation shortcuts and toggle recording mode

### LINKED ISSUE #4: Feature request: Toggle dictation with Fn+Space (press to start, press to stop)
**Problem**

Currently dictation requires holding down the Fn key for the duration of the transcription. For longer dictation sessions or conversations, this is uncomfortable and impractical.

**Proposed Solution**

Add a toggle mode where pressing Fn+Space (or a configurable hotkey) starts dictation, and pressing it again stops dictation and finalises the transcription. No need to hold anything down, just tap to start, talk, tap to stop.

**Why This Is Useful**

Hold-to-dictate works for quick snippets, but for anything longer than a sentence or two it's ergonomically painful. A toggle mode would make FreeFlow viable for longer dictation, emails, meeting notes, and hands-free workflows.

P.S. Thanks for making this, lovely work

### LINKED ISSUE #10: Add hotkey delay before starting dictation
Hello

I have my right option key set as my push-to-talk hotkey. Sometimes I accidentally hit it, which starts dictation instantly. I think it would be nice to have a bit of a delay (2 seconds? Or maybe configurable?) to avoid this. So, hold the hotkey for N seconds before dictation begins. 

Thanks

### LINKED ISSUE #25: Need to be able to change the keyboard mapping.
I think when you click on the app, it should open something up that can let you change the keyboard mapping or add to the dictionary.

### LINKED ISSUE #29: Add a configurable push-to-record delay to differentiate between pressing and holding

marcbodea:
## Summary

This PR adds configurable dictation shortcuts and introduces a toggle-based recording mode alongside the existing hold-to-record workflow.

Users can now choose separate shortcuts for:

- hold to record
- tap to start/stop dictation

Either workflow can also be disabled entirely, as long as one shortcut remains enabled.

It also includes a follow-up fix for custom shortcut handling so combinations like `Option + \`` are less likely to leak characters into the focused text field and are more reliable when pasting the final transcript.

## What Changed

- replaced the single hotkey model with separate hold and toggle shortcut bindings
- added mode-aware shortcut session handling for:
  - hold-to-record
  - tap-to-toggle
  - switching from hold mode into toggle mode mid-session
- added the ability to disable hold-to-talk or tap-to-toggle entirely
- added shortcut configuration UI in setup and settings
- updated menu bar and onboarding copy to reflect the new shortcut model
- improved custom hotkey handling so shortcut keystrokes are less likely to appear in the active text field
- delayed synthetic paste until shortcut-related keys are released
- simplified the recording overlay by removing the hold/tap mode labels and keeping only the stop control in toggle mode

## Small UI / Onboarding Changes

- adjusted onboarding layout to better fit the larger shortcut configuration content
- moved the setup status/progress UI to make the flow fit more cleanly
- updated setup and menu copy to adapt when only one shortcut mode is enabled

## Why

The previous model required users to hold a single key for the full duration of dictation. That works for short utterances, but it is awkward for longer sessions and doesn’t support users who want a different keyboard mapping.

This change keeps the existing quick hold-to-talk workflow while adding a more flexible toggle flow, configurable shortcuts, and the ability to use only the interaction model each user prefers.

## Screenshots

<p align="center">
  Shortcut settings
  Onboarding shortcut step
  Onboarding toggle shortcut step
  Recording overlay
</p>

## Issues

Closes #4  
Closes #25
Closes #10  
Closes #29

## Testing

- verified hold-to-record still starts on key down and stops on release
- verified toggle mode starts on first press and stops on second press
- verified switching from hold mode into toggle mode works
- verified hold-only configuration works
- verified tap-only configuration works
- verified the app prevents disabling both shortcuts at the same time
- verified custom shortcuts can be captured and saved from setup/settings
- verified custom combinations like `Option + \`` behave more reliably and no longer leak the shortcut character as often

### DIFF
--- a/Sources/HotkeyManager.swift
+++ b/Sources/HotkeyManager.swift
@@ -142,33 +142,74 @@ final class HotkeyManager {
     }
 
     private func handleFlagsChanged(_ event: NSEvent) -> Bool {
-        let shouldConsume = shortcutReferencesModifierKeyCode(event.keyCode)
+        let shouldConsumeBefore = shouldConsumeModifierEvent(
+            for: event.keyCode,
+            pressedKeys: pressedKeyCodes,
+            pressedModifiers: pressedModifierKeyCodes
+        )
+
         if ShortcutBinding.modifierKeyCodes.contains(event.keyCode) {
-            if pressedModifierKeyCodes.contains(event.keyCode) {
-                pressedModifierKeyCodes.remove(event.keyCode)
+            var updatedModifierKeyCodes = pressedModifierKeyCodes
+            if updatedModifierKeyCodes.contains(event.keyCode) {
+                updatedModifierKeyCodes.remove(event.keyCode)
             } else {
-                pressedModifierKeyCodes.insert(event.keyCode)
+                updatedModifierKeyCodes.insert(event.keyCode)
             }
+            pressedModifierKeyCodes = updatedModifierKeyCodes
         }
+
+        let shouldConsumeAfter = shouldConsumeModifierEvent(
+            for: event.keyCode,
+            pressedKeys: pressedKeyCodes,
+            pressedModifiers: pressedModifierKeyCodes
+        )
         evaluateActiveBindings()
-        return shouldConsume
+        return shouldConsumeBefore || shouldConsumeAfter
     }
 
     private func handleKeyDown(_ event: NSEvent) -> Bool {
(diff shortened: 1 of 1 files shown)


marcbodea:
My pleasure, give me 5 minutes will commit also option to disable either hold or tap hotkeys.

PR #36, opened 2026-03-15
feat: OpenAI API Option And Model Use Without Using Custom URL

ChandlerVS:
# OpenAI API Option Without Using Custom URL

This PR allows for the selection and use of OpenAI's API without having to input custom URLs, etc. It also uses openai models instead. I think this solution is nice because you can switch between OpenAI and Groq witout tinkering with URL settings, etc.

Users can select from Groq, OpenAI, or Custom during setup:

Screenshot 2026-03-14 at 12 41 13 PM

If OpenAPI is selected, it shows a similar input screen to Groq, but with links to OpenAI instead:

Screenshot 2026-03-14 at 12 41 15 PM

It is also configurable in settings after setup:

Screenshot 2026-03-14 at 12 40 47 PM

## Additional Notes

- Groq is the fallback, so existing users should not be affected
- Saving an API key does not override the API key of a different provider, so you can freely switch between Groq and OpenAI without hassle
- When selecting OpenAI, an OpenAI model is used for chat instead of llama
- I may have missed it, but I did not see Whisper 3 as an option within OpenAI's API itself. It seems the only available Whisper model is ``whisper-1`` (https://developers.openai.com/api/docs/guides/speech-to-text). I am unsure if the 4o models from the doc will work well here.

### DIFF
--- a/Sources/APIProvider.swift
+++ b/Sources/APIProvider.swift
@@ -34,7 +34,7 @@ enum APIProvider: String, CaseIterable, Identifiable {
     var chatModel: String {
         switch self {
         case .groq: return "meta-llama/llama-4-scout-17b-16e-instruct"
-        case .openai: return "gpt-4o-mini"
+        case .openai: return "gpt-5-mini-2025-08-07"
         case .custom: return "meta-llama/llama-4-scout-17b-16e-instruct"
         }
     }


ChandlerVS:
Closing in support of #43.

PR #40, opened 2026-03-20
Add elapsed recording timer with auto-stop and configurable max duration

xROBERTINOx:
## Summary
- Added a live elapsed timer (M:SS) displayed next to the waveform in the recording overlay
- Timer changes color as it approaches the limit (white → yellow at 30s remaining → red at 10s remaining)
- Recording auto-stops and transcribes when the time limit is reached (default: 2 minutes)
- Added a "Recording Limit" card in Settings → General where users can keep the default (120s) or set a custom duration (15–600s)
- Includes an info note warning that longer recordings take more time to transcribe

## Files Modified
- `Sources/RecordingOverlay.swift` — Added `ElapsedTimerView`, `elapsedSeconds`/`maxRecordingSeconds` on overlay state, widened overlay panel to fit timer
- `Sources/AppState.swift` — Added `maxRecordingSeconds` as a persisted setting, recording timer with auto-stop logic
- `Sources/SettingsView.swift` — Added "Recording Limit" settings card with default/custom options

## Why
Groq's Whisper transcription has a 20-second timeout. Uncompressed PCM audio at ~10 MB/min means recordings over ~3 minutes risk timing out. The 2-minute default provides a safe margin while covering typical dictation use cases.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -51,6 +51,8 @@ final class AppState: ObservableObject, @unchecked Sendable {
     private let forceHTTP2TranscriptionStorageKey = "force_http2_transcription"
     private let transcribingIndicatorDelay: TimeInterval = 1.0
     let maxPipelineHistoryCount = 20
+    let maxRecordingSeconds: Int = 120
+    private var recordingTimer: Timer?
 
     @Published var hasCompletedSetup: Bool {
         didSet {
@@ -785,6 +787,7 @@ final class AppState: ObservableObject, @unchecked Sendable {
                         .sink { [weak self] level in
                             self?.overlayManager.updateAudioLevel(level)
                         }
+                    self.startRecordingTimer()
                 }
             } catch {
                 DispatchQueue.main.async {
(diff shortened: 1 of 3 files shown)


marcbodea:
> Can you send a video showing this working? Thank you! Also FYI that it no longer uses uncompressed audio. Audio is now compressed.

Since this was merged https://github.com/zachlatta/freeflow/pull/38, the theoretical time limit for a transcription to fit in groq's 25MB limit is around 12 minutes. I've ran a couple of tests on audio files with that length and it seems the transcription finishes in around 3-4 seconds, so the 20 second timeout should be plenty.

I would say this covers most dictation use cases, what would maybe make sense is a countdown timer when getting close to the limit (instead of permanent), but merging https://github.com/zachlatta/freeflow/pull/43 complicates things significantly, making it hard to set good defaults based on provider/plan/etc. so it would be up to the user to manually set the time limit.

zachlatta:
Closing this due to no response. Would be happy if you re-opened it!

PR #67, opened 2026-04-08
Change default processing to gpt-oss-20b with fallback to llama-4-scout

marcbodea:
This changes the default post-processing model from llama-4-scout to gpt-oss-20b.

## Why:

gpt-oss-20b performs materially better in our prompt evals, especially on cleanup quality and anti-hallucination behavior.
It also appears to handle non-Latin dictation better, including cases like Japanese where we were more likely to get empty output.
Because the free-plan token limit for gpt-oss-20b is lower than Llama, this also adds an automatic fallback to llama-4-scout when post-processing hits HTTP 429 rate-limit responses. That keeps the higher-quality default while reducing the risk of degraded behavior once users exhaust the GPT daily cap.

## Scope:

Context model remains unchanged.
Only the post-processing default is updated.
Fallback is server-side and automatic; no UI changes.

Closes #64

### DIFF
--- a/Sources/PostProcessingService.swift
+++ b/Sources/PostProcessingService.swift
@@ -88,7 +88,8 @@ Output hygiene:
 
     private let apiKey: String
     private let baseURL: String
-    private let defaultModel = "meta-llama/llama-4-scout-17b-16e-instruct"
+    private let defaultModel = "openai/gpt-oss-20b"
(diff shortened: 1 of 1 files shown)


zachlatta:
Please see #69 for why I reverted this.

Issue #69, opened 2026-04-09
New version unusable and Issues with long transcriptions


prashstarfish:
With the new update, which is: FreeFlow build-20260408-145854-edc4ede. 
The GPT-OSS 20 billion was added and that has created a bug in the app where, when I am speaking for longer durations—more than a minute—it doesn't give any result. So you have to go into the settings, go to the run log, and then copy the transcription from there.

prashstarfish:
The other thing is that the new update with the new model has completely made the app useless because instead of transcribing, it is now paraphrasing, which is awful.

zachlatta:
@prashstarfish: Thanks so much for reporting this, I've just reverted the change. Please go to Settings > Update to update to the new version that fixes that - thank you!

@marcbodea: Can you investigate this and submit a new PR that doesn't cause this problem? Thanks!

marcbodea:
On it, thanks for reporting! Not sure if the issue is caused more by #67 or #49 

If it's not too much of a hassle, could you paste the full prompt from the run log here so I can test properly? @prashstarfish

zachlatta:
I've just reverted both #67 and #49 out of an abundance of caution. As soon as we get to the root of this and have a solution, I'm eager to switch back to a better model and better system prompt

marcbodea:
Great, thanks.

It seems it's a very stupid issue caused by groq applying a 1024 token completion limit by default when not specified in the request **for reasoning models**. Specifying a 4096 token limit in the request solves the issue.

I will push a PR now that:
1. adds the 4096 token limit for gpt-oss-20b
2. if the transcript from gpt-oss-20b is empty, fallback to llama
3. if transcript from llama is empty, paste raw_transcript

The paraphrasing issue is most likely connected to the new prompt, it would be best to have the original raw transcript that was paraphrased to be able to test.

To solve this, I would include a toggle in the settings to allow users to choose between the highly-guided prompt that was in the previous PR, or the original low-guidance prompt that more unlikely to paraphrase. What do you think?

zachlatta:
I would prefer to have 1 sane default instead of toggleable stuff. That'll make it easier for us to identify and fix issues in the future. How does that sound to you? I like your proposed solution

--
Zach Latta

https://zachlatta.com ( http://zachlatta.com )
@zachlatta ( https://twitter.com/zachlatta ) / fb ( https://www.facebook.com/crynix ) / github ( https://github.com/zachlatta )

marcbodea:
[...]

PR #73, opened 2026-04-11
Fix/postprocess fallback

marcbodea:
This PR reintroduces the post-processing changes needed to fix blank output on long dictations (#69)

It updates PostProcessingService to use openai/gpt-oss-20b by default with meta-llama/llama-4-scout-17b-16e-instruct as a fallback, sets max_completion_tokens: 4096 for GPT-OSS only, and retries on Llama if GPT-OSS returns empty output. If both post-processing attempts fail, the app falls back to the raw transcript instead of pasting blank text.

It also adds a README note with the original prompt for users who want a more literal, less context-aware cleanup style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,32 @@ One of the cool features is that it's context aware. If you're replying to an em
 
 An added bonus is that there's no FreeFlow server, so no data is stored or retained - making it more privacy friendly than the SaaS apps. The only information that leaves your computer are the API calls to Groq's transcription and LLM API (LLM is for post-processing the transcription to adapt to context).
 
+If you prefer less context fitting, you can paste the original prompt from `main` into the custom system prompt setting:
+
+<details>
+  <summary>If you prefer less context fitting</summary>
+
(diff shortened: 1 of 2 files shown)


zachlatta:
Can you confirm you've manually tested this before I merge? You've submitted a bunch of great PRs before so I trust you, but I just want to double check

marcbodea:
Yes, I also built an adapter that runs the app's processing pipeline with audio files for testing. I have not included it in the PR to avoid bloat but can do if you prefer.

Just tested with a 4 minute dictation through the app itself and it works with no problems

zachlatta:
Thank you so much! I'm not sure what the best approach will be, but it'd be good for us to have a clear test / benchmark suite in the repo for the future as the app gets more complex

marcbodea:
My pleasure. I'll open an issue to discuss testing. 

As you said not sure what the best approach is, as running live groq api queries in CI doesn't sound right but could work

Issue #68, opened 2026-04-09
Filter out rubbish text outputted from empty speech audio


thsunkid:
Whisper v3 is well-known for outputting garbage like "thank you", "thank you very much" when the audio is mostly noise or with empty speech.
There is a trick to handle this without false positive (i.e. when the user actually says 'thank you'). Want me to send a PR?

zachlatta:
Yes, please!

m13v:
this is a really common whisper v3 issue. the hallucination on silence is well documented and the 'thank you' / 'thanks for watching' outputs happen because those phrases dominate the training data endings.

the trick that works without false positives: run Silero VAD on the audio chunk before sending to whisper. if the VAD speech probability for the entire chunk is below a threshold (we use 0.4), skip transcription entirely. this catches the silence/noise case without filtering out someone actually saying 'thank you' because real speech has much higher VAD probability (typically 0.85+).

another complementary approach: compute the audio energy (RMS) and if it's below a floor (like -40dB), don't even run VAD. this saves the VAD inference cost on truly silent chunks.

m13v:
our transcription pipeline implements the VAD gating approach with audio energy thresholds before inference: https://github.com/m13v/fazm/blob/main/Desktop/Sources/TranscriptionService.swift

marcbodea:
It seems whisper through groq returns a "No Speech Probability" field which we might be able to use very easily. Will run some tests and see how it performs.

m13v:
nice, that field is solid for this. in our experience the no_speech_prob threshold sweet spot is around 0.6 to 0.8 for filtering hallucinations without dropping real speech. below 0.6 you start letting some garbage through, above 0.8 you might clip quiet or hesitant speakers.

one thing to watch for: the field can occasionally report low no_speech_prob (meaning "speech detected") on repetitive background noise like fans or typing. if that becomes an issue, combining it with a simple RMS energy check as a first pass catches those cases cheaply.

marcbodea:
Thanks for the insight! Hope this way we'll get most of the benefits of VAD gating without spinning up much extra, RSM check should be fairly cheap.

thsunkid:
Here's the PR @zachlatta!
https://github.com/zachlatta/freeflow/pull/81

> no_speech_prob threshold sweet spot is around 0.6 to 0.8

sounds wrong to me. no_speech_prob score for audio with any speech at all (including background speech) is often very low (<0.02). curious if you have tested that on any dataset @m13v ?

PR #81, opened 2026-04-14
Filter 'thank you' hallucinations using whisper no_speech_prob

thsunkid:
Issue: https://github.com/zachlatta/freeflow/issues/68

`whisper-large-v3` model often outputs random phrases for the audio with no speech, especially "thank you" / "thank you very much". Using `no_speech_prob` value from the model, we can filter these cases out.

## Changes

- Request `response_format=verbose_json` from Groq so the response includes `segments[].no_speech_prob`.
- If the normalized text matches a known hallucination phrase (2 phrases right now, but can expand) *and* `no_speech_prob` clears that phrase's threshold, return empty.
- The current threshold `0.06` is fine-tuned on a private dataset (size ~500) recorded from different environments  (quiet + noisy, real speech + silence). Real speech that say 'thank you' clustered well below it (<0.02)

## Summary by CodeRabbit

* **New Features**
  * Intelligent hallucination detection now automatically identifies and filters out false or incorrectly generated transcriptions, improving overall accuracy and reliability
  * Enhanced transcription responses now utilize verbose JSON format with more comprehensive and structured data for detailed analysis and processing

### DIFF
--- a/Sources/TranscriptionService.swift
+++ b/Sources/TranscriptionService.swift
@@ -9,6 +9,7 @@ class TranscriptionService {
     private let baseURL: String
     private let forceHTTP2: Bool
     private let transcriptionModel = "whisper-large-v3"
+    private let transcriptionResponseFormat = "verbose_json"
     private let transcriptionTimeoutSeconds: TimeInterval = 20
     private let uploadSampleRate = 16_000.0
     private let uploadChannelCount: AVAudioChannelCount = 1
@@ -83,6 +84,7 @@ class TranscriptionService {
             audioData: audioData,
             fileName: fileURL.lastPathComponent,
             model: transcriptionModel,
+            responseFormat: transcriptionResponseFormat,
             boundary: boundary
         )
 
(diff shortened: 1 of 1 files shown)


marcbodea:
In general I would avoid hard-coding phrases and rely on the no_speech_prob.

I think the right and "clean" approach, which works for longer dictations with potential pauses in the middle would be to process each segment individually and drop them if no_speech_prob is higher than x.

From my testing, I have not found many cases where there is no speech and the prob is lower than 0.5, much less 0.06. Can you share the dataset you used?

thsunkid:
@marcbodea 

I can't share the entire dataset but here is a couple bad 'thank you' examples, whose audio file name is the `no_speech_prob`
[bad_thank_you.zip](https://github.com/user-attachments/files/26705502/bad_thank_you.zip)

I can't find the example that is exactly at `0.06`, but pretty sure that I have encountered them in the past when I tuned the threshold.

We can always start with a high threshold, as you suggested. But eventually we will want to improve recall without losing precision. Then we’ll need to capture the model’s failure mode more precisely, thus the hard-coded phrases approach.

marcbodea:
Thanks for sharing, makes sense.

I guess the only risk with your solution is a dictated "thank you" in a sub-optimal environment that gets a relatively high no_speech_prob.

What do you think about merging with a 0.1 threshold and see what the experience is?

thsunkid:
> I guess the only risk with your solution is a dictated "thank you" in a sub-optimal environment that gets a relatively high no_speech_prob.

Yup. The risk is much lower now that we only apply it to this transcript output. But also from my testing, the audio with actual 'thank you' speech has very low no speech prob.

0.1 makes sense. Will commit it shortly

thsunkid:
thanks for finishing it @marcbodea  :)

PR #95, opened 2026-04-16
Swap transcript sanitizers in post-processing flow

marcbodea:
## Summary
- Swapped the sanitization functions used by the post-processing and command-mode transcript paths.
- This restores each path to using its intended sanitizer before empty-output validation.
- The change is limited to `Sources/PostProcessingService.swift`.

## Testing
- Not run

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed post-response sanitization so each processing path now uses the correct cleanup and empty-output checks.
  * Ensures consistent trimming, quote handling, and empty-result normalization across modes to prevent spurious empty or quoted transcripts.

### DIFF
--- a/Sources/PostProcessingService.swift
+++ b/Sources/PostProcessingService.swift
@@ -380,7 +380,7 @@ Model: \(model)
             throw PostProcessingError.invalidResponse("Missing choices[0].message.content")
         }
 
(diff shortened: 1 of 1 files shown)


marcbodea:
@coderabbitai review

PR #102, opened 2026-04-20
feat: add Output Language setting for automatic translation

marwei:
## Summary
- Adds an **Output Language** picker in Settings > General (English, Chinese, Spanish, French, Japanese, Korean, German, Portuguese)
- When set, the post-processing system prompt is modified to translate transcribed speech into the selected language
- Works for both dictation mode and edit/command mode

## Test plan
- [x] Open Settings > General, confirm "Output Language" card appears after API Key
- [x] Set to "Spanish", speak English, verify output is in Spanish
- [x] Set to "Same as spoken", speak English, verify output stays in English
- [x] Check Run Log to confirm the modified system prompt includes translation instructions
- [x] Test edit mode with output language set — verify transformed text is in target language

## Summary by CodeRabbit

* **New Features**
  * Added an "Output Language" option in General Settings with a "Same as spoken" choice plus multiple languages.
  * Selected preference is saved persistently and restored across launches.
  * The chosen output language is applied to transcription post-processing, including retry flows, and to command/edit transformations so results are produced in the selected language.


marcbodea:
Hi, the idea of having translation built in is great. I would like to discuss the implementation in more detail:

For edit mode, not sure this is needed as the speaker can just say "translate this to x language" and it seems to work well from my testing. Have you encountered any issues with that?

I am hesitant to add the persistent settings toggle as I am not sure what the use case is for having all dictations translated. What do you think about adding a "trigger translation" phrase that if found in the transcription will add the "translate this to x language" to the prompt? I will test a couple of runs to see if having it in the user prompt is enough.

In general I think it's better to not dynamically modify the system prompt between runs as that would bust the cache on Groq and cause extra usage.

marwei:
Thanks for taking the time to dig into this — these are fair concerns.

I see this primarily as a UX improvement for non-native English speakers, a user persona underserved by the trigger-phrase alternative.

Consider a non-native English speaker who uses Flow throughout their workday — Slack messages, emails, PR descriptions, Linear tickets, docs. Their native language might be Mandarin, Spanish, Portuguese, German, Hindi, whatever. Their output language, however, is English, essentially 100% of the time in a work context. For this user, translation isn't an occasional intent they opt into per-utterance; it's the stable default of their entire workflow. You're right that "translate this to X" works well when someone consciously wants a translation for a specific utterance — but asking this user to prepend it to every single dictation, or to remember a trigger phrase each time, turns what should be a fluent dictation experience into a constant cognitive tax, and worse, a constant reminder that the tool wasn't really built with them in mind. A trigger phrase solves the mechanics but not the ergonomics.

This is actually one of the higher-leverage UX improvements Flow could make for non-English-speaking users, and a real lever for adoption beyond the native-English audience. The population of knowledge workers worldwide who think in one language and write professionally in another is enormous, and right now Flow implicitly assumes input language == output language. A persistent setting is the simplest, most honest way to say "your output language is a preference, not a per-utterance decision."

On the Groq caching concern specifically: because this is a setting rather than a dynamic per-run value, the system prompt stays stable until the user explicitly changes it in settings. Cache invalidation would only happen on setting changes, which should be rare in practice. Happy to verify this empirically if helpful, but I don't think it meaningfully degrades cache hit rate for any given user.

marcbodea:
Makes sense. My only worry left is that the models that can realistically be used for post-processing are not that great at translation. Do you have experience using this feature in practice and have seen good results? I couldn't really find some good benchmarks around but would love seeing some.

Re: groq caching should be OK with static setting

Thanks!

marwei:
Thanks — ran an eval in both directions to check. 280 samples across 14 language pairs (20 each), using a script that mirrors `PostProcessingService.postProcess()` + `applyOutputLanguage` verbatim and hits the real Groq API. Inputs are synthesized but representative of typical Flow-user messages (Slack/email/PR/ticket/meeting-note/code-review).

**Headline**: 12 of 14 pairs are very strong, and the remaining two (English → CJK) are still solid on translation quality itself.

The issues I saw on en-ja and en-zh weren't translation-quality problems — they were cases where `llama-4-scout` ignored the directive entirely and echoed the English back (5/20 en-ja, 3/20 en-zh). The same model has directive-compliance slips on plain English→English dictation too, sometimes returning `"Here's the cleaned transcript:"`-style preamble. So this is really a llama-scout reliability issue with adhering to system-prompt directives in general, not a translation-quality issue. When it does translate, the output is clean and idiomatic.

Couldn't get a full run on `gpt-oss-20b` — it hit Groq's daily token quota within the first few requests on two separate keys. Given how early that burns, most real user calls end up on `llama-4-scout` anyway, so these results probably reflect what most users actually see.

Caveats: N=20/pair, synthesized inputs, single LLM judge.

Script + raw 280-sample results: https://gist.github.com/marwei/3d7efb7126bea19831d22465d8bf936d

marcbodea:
Look good! Thank you for putting this together.

Will solve the conflicts and merge shortly.

PR #103, opened 2026-04-21
Add realtime streaming transcription support and reduce perceived latency on long dictations

iris-sfg:
This MR adds realtime streaming transcription support to FreeFlow.

Instead of waiting for a full recorded segment to finish before starting speech-to-text, FreeFlow can now send audio incrementally to any OpenAI Realtime API compatible transcription backend and consume partial transcript deltas as they arrive.

The main product benefit is lower response time, especially on long dictations. Text can now be processed during the dictation itself instead of only after the speaker stops, which reduces perceived latency and makes the experience feel much more live.

This also makes FreeFlow compatible with local realtime transcription backends, including the new `mlx-audio` Voxtral Realtime path:
- `https://github.com/Blaizzy/mlx-audio/pull/661`

In practice, this branch has been tested against a local realtime setup and works end-to-end with incremental audio input and partial transcript updates.

## What changed

- add realtime streaming transcription flow in FreeFlow
- send audio progressively instead of waiting for a fully buffered transcription request
- consume and surface partial transcript deltas as they arrive
- keep compatibility with OpenAI Realtime API style backends

## Why this matters

- improves the UX on long utterances because transcription starts before the end of speech
- makes local low-latency transcription backends usable in FreeFlow

## Validation

- tested locally end-to-end with a realtime transcription backend
- verified behavior on longer dictation flows where streaming meaningfully reduces response time

## Summary by CodeRabbit

* **New Features**
  * Real-time transcription during recording with live partial updates and seamless fallback to file-based transcription.
  * Settings UI to enter and persist a transcription API URL, API key, and realtime model, with Clear buttons and trimmed input.
  * Toggle to enable/disable realtime streaming.
  * Improved audio capture that streams converted PCM16 audio for realtime transcription.

* **Bug Fixes**
  * More reliable realtime teardown and fallback handling to avoid stuck or duplicated transcription attempts.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -1729,7 +1729,11 @@ final class AppState: ObservableObject, @unchecked Sendable {
         fileURL: URL
     ) async throws -> String {
         if let realtimeService {
-            return try await realtimeService.commitAndAwaitFinal()
+            do {
+                return try await realtimeService.commitAndAwaitFinal()
+            } catch {
+                return try await fileService.transcribe(fileURL: fileURL)
+            }
         }
         return try await fileService.transcribe(fileURL: fileURL)
     }
@@ -1806,6 +1810,9 @@ final class AppState: ObservableObject, @unchecked Sendable {
             self.audioRecorder.onPCM16Samples = nil
             self.transcriptionTask?.cancel()
             self.transcriptionTask = Task {
+                defer {
+                    activeRealtime?.cancel()
+                }
                 do {
                     let transcriptionService = try TranscriptionService(
                         apiKey: self.resolvedTranscriptionAPIKey,
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -1966,7 +1966,10 @@ final class AppState: ObservableObject, @unchecked Sendable {
         if let realtimeService {
             do {
                 return try await realtimeService.commitAndAwaitFinal()
+            } catch is CancellationError {
+                throw CancellationError()
             } catch {
+                try Task.checkCancellation()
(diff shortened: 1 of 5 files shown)


marcbodea:
uuu this is very interesting, thanks!

iris-sfg:
I'm using it with mlx-audio [server](https://github.com/Blaizzy/mlx-audio/pull/661). Using streamed transcription (and voxtral realtime model that is better than parakeet-v3), allows my laptop to process both context processing and transcription in parallel (no MLX concurrency issues), maximizing compute usage and minimizing latency by launching post-processing as soon as dictation finishes.

marcbodea:
Added a new field to the provider settings to allow the user to enter a streaming transcription model, since OpenAI's API requires it to be specified.

iris-sfg:
> Added a new field to the provider settings to allow the user to enter a streaming transcription model, since OpenAI's API requires it to be specified.

Since we are already using the same `Transcription API URL`for both mode, I figured we could use the same `Transcription Model` setting for both mode too.

marcbodea:
> > Added a new field to the provider settings to allow the user to enter a streaming transcription model, since OpenAI's API requires it to be specified.
> 
> Since we are already using the same `Transcription API URL`for both mode, I figured we could use the same `Transcription Model` setting for both mode too.

That could work as well, I figured some models do not support streaming transcription, so it makes it easier to manage the fallback. In the original implementation, the model was not being sent to the API at all.

Feel free to revert it to use the transcription model instead of adding a new field and then i can merge

Thanks

iris-sfg:
You are right, let's keep it like this.

PR #104, opened 2026-04-21
Edit Mode: clipboard fallback for non-AX apps (Electron/Chromium)

marwei:
## Summary
- Edit Mode currently only reads the selection via the Accessibility API, which returns nothing in Electron/Chromium apps (VS Code, Slack, Discord, Chrome web content). The intent resolver then silently falls through to dictation, so selected text gets **replaced** instead of transformed.
- This PR adds a ⌘C fallback: when AX returns no selection and Edit Mode is enabled, synthesize ⌘C, read the pasteboard, and restore the prior clipboard. `HotkeyManager` tags self-injected events with a `CGEventSource.userData` sentinel and drops them in its event tap, so synthetic ⌘C/⌘V can't re-trigger the dictation shortcut.
- New **"Use clipboard fallback to read selection"** toggle inside Edit Mode settings (default on). Only runs when Edit Mode is enabled + (automatic style OR manual style with modifier held), so the clipboard is never touched for plain dictation users.

## Test plan
- [x] Build via \`make\`, ad-hoc sign if needed
- [x] TextEdit: select text → Edit Mode transforms selection (AX fast path unchanged)
- [x] VS Code / Slack: select text → Edit Mode transforms selection (previously replaced)
- [x] Copy something, run Edit Mode on a different selection, then ⌘V elsewhere — original clipboard still intact
- [x] Toggle the new setting off → Edit Mode behavior reverts to current (replace-on-Electron)
- [x] Dictation shortcut bound with ⌘ modifier: Edit Mode in VS Code doesn't re-trigger the shortcut

## Release Notes

* **New Features**
  * Added clipboard fallback option in Settings under command mode. When enabled, provides an alternative method to capture highlighted text when standard accessibility methods are unavailable.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -187,6 +187,7 @@ final class AppState: ObservableObject, @unchecked Sendable {
     private let commandModeEnabledStorageKey = "command_mode_enabled"
     private let commandModeStyleStorageKey = "command_mode_style"
     private let commandModeManualModifierStorageKey = "command_mode_manual_modifier"
+    private let useCopyFallbackForSelectionStorageKey = "use_copy_fallback_for_selection"
     private let transcribingIndicatorDelay: TimeInterval = 0.25
     private let clipboardRestoreDelay: TimeInterval = 0.15
     let maxPipelineHistoryCount = 20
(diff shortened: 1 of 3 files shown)


marcbodea:
I've been using the feature in Slack and Chrome for a while now without issues. I did just merge a PR that might be related to this. Could you update and re-test?
If you still have the issue let's explore further.

marwei:
Thanks for the heads up! I merged main into this branch and re-tested. Edit Mode is now working correctly in Slack, Chrome, and VS Code without the clipboard fallback — your recent PR resolved the underlying issue.

Closing this since the fallback is no longer needed.

PR #105, opened 2026-04-21
Stop Fn-bound shortcut from triggering on arrow keys

baslefeber:
## Summary

On macOS, `NSEvent.ModifierFlags.function` is set automatically for arrow keys, F-keys, and navigation keys (Home/End/Page Up/Page Down/Forward Delete) regardless of whether the physical Fn key is held. The shortcut matcher derived its pressed-modifier snapshot from that flag on every `keyDown`/`keyUp`, so if the user had configured a Fn-based hold or toggle binding (the default), pressing any arrow key would falsely activate the binding and start recording.

## Repro (before this PR)

1. Keep default shortcut (hold Fn to talk) or set any binding that references Fn.
2. Press any arrow key (←, →, ↑, ↓) in any text field.
3. FreeFlow starts a recording session.

Same for Home/End/Page Up/Page Down/Forward Delete and the F-row keys.

## Fix

- Track the physical Fn state inside each backend using `flagsChanged` events for keyCode 63 — those events are authoritative.
- Pass that trusted state into `ModifierKeyEventState.pressedModifierKeyCodes(for:trustedFunctionKeyIsDown:)` when building the snapshot, so the `.function` flag on arrow/nav/F-key events is ignored in favor of the backend's real Fn state.
- Reset the tracked Fn state on backend reset.

`LocalShortcutCaptureBackend` already maintained its own `pressedModifierKeyCodes` set from `flagsChanged`, so it simply reuses that. `GlobalShortcutBackend` gains a small `fnKeyIsDown` field.

## Test plan

- [x] `swiftc -typecheck` clean on all Swift sources
- [x] Full `swiftc` build produces a binary
- [ ] With Fn-hold binding: pressing ←/→/↑/↓ does not start a recording
- [ ] With Fn-hold binding: holding Fn still starts a recording
- [ ] With Fn-hold binding: holding Fn then pressing an arrow still keeps the recording session active (trusted Fn state preserved across the snapshot)
- [ ] Cmd+Fn toggle still works

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Function (Fn) key detection and tracking for more accurate modifier recognition when recording and detecting shortcuts.
  * Fn state is now initialized and re-evaluated when capture starts or resumes, included in modifier snapshots, and reliably reset on backend reset to reduce incorrect shortcut interpretations.

### DIFF
--- a/Sources/GlobalShortcutBackend.swift
+++ b/Sources/GlobalShortcutBackend.swift
@@ -28,6 +28,7 @@ final class GlobalShortcutBackend {
     func start() throws {
         stop()
         try installEventTap()
+        fnKeyIsDown = ModifierKeyEventState.currentFunctionKeyIsDown()
     }
 
     func stop() {
@@ -104,6 +105,7 @@ final class GlobalShortcutBackend {
             notifyBackendReset()
             if let tap = eventTap {
                 CGEvent.tapEnable(tap: tap, enable: true)
+                fnKeyIsDown = ModifierKeyEventState.currentFunctionKeyIsDown()
             }
             return Unmanaged.passUnretained(event)
 
--- a/Sources/GlobalShortcutBackend.swift
+++ b/Sources/GlobalShortcutBackend.swift
@@ -20,6 +20,7 @@ enum GlobalShortcutBackendError: LocalizedError {
 final class GlobalShortcutBackend {
     private var eventTap: CFMachPort?
     private var eventTapRunLoopSource: CFRunLoopSource?
+    private var fnKeyIsDown = false
 
     var onInputEvent: ((ShortcutInputEvent) -> ShortcutConsumeDecision)?
     var onEscapeKeyPressed: (() -> Bool)?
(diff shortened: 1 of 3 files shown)


marcbodea:
I merged a PR a few hours ago that changed the way shortcuts are handled significantly. Can you update and check if you're still encountering this issue? From my testing it seems to be fixed.

baslefeber:
Hey @marcbodea, thanks! Just built from latest main (1167337, 0.3.0) to double-check and it still reproduces for me. Any arrow key in a text field starts a recording.

Looking at the commits since #98 I don't see anything touching the files affected (ModifierKeyEventState, GlobalShortcutBackend, LocalShortcutCaptureBackend), so I think this fix is still needed.

The root cause is that macOS auto-sets NSEvent.ModifierFlags.function on those keys even when fn isn't actually held, and the modifier snapshot in handleKeyDown then picks that up as fn being pressed.

Could be a keyboard or setup difference on your side — I'm on Apple Silicon, default fn-hold shortcut

marcbodea:
lgtm

PR #114, opened 2026-04-22
Show hint when Preview is clicked while system audio is muted

### LINKED ISSUE #106: Alert when system volume is muted on feedback sound preview
Big fan of FreeFlow. The feedback sound volume control has been super helpful for me. I thought Wispr Flow's sounds were a little quiet, so being able to turn these up is a win.

One small UX nitpick:

When I clicked the preview button in Settings to hear what the feedback sound would sound like, nothing happened. I sat there clicking it a few times wondering if it was broken, and eventually realized my Mac's system volume was muted. Not the app's fault at all, but a quick "hey your system is muted" hint would have saved me a couple minutes of confusion.

## Suggestion

When the preview button is clicked, check the system output mute state (and maybe volume level). If muted or near-zero, show an inline hint like:

> System is muted. Unmute to hear the preview.

Placement could be right next to the button so it is obvious.

Would love to see this added.

ojhurst:
Closes #106.

## What

When the **Preview** button in Settings > Sound Volume is clicked while the system is muted (or volume is below 10%), an inline hint appears next to the button:

> 🔇 System volume is muted or very low. Unmute to hear the preview.

This matches the suggestion in the issue — the button appearing to do nothing when system audio is muted was confusing. Now the user gets immediate feedback.

## How

- New `SystemAudioStatus` enum namespace in `Sources/SystemAudioStatus.swift` with two Core Audio helpers: `isDefaultOutputMuted()` and `defaultOutputVolume()`.
- Both fail gracefully (return `false` / `nil`) if a property can't be read, so the hint never falsely appears on devices that don't expose mute/volume (HDMI, some Bluetooth setups).
- `defaultOutputVolume()` falls back to reading channels 1 and 2 if the main element doesn't expose volume — common on many external output devices.
- Threshold: 10%. macOS volume steps are ~6.25%, so the first tick above mute (6.25%) triggers the hint, but the second tick (12.5%) does not.

## Testing

Tested on macOS with built-in speakers. All combinations behave correctly:

| System state | Click Preview | Result |
|---|---|---|
| Muted | ✓ | Hint appears |
| Volume 1 notch up (below 10%) | ✓ | Hint appears |
| Volume well above 10% | ✓ | No hint, sound plays |
| Back to muted | ✓ | Hint appears again |
| Toggle mute + various volume levels | ✓ | All correct |

## Summary by CodeRabbit

* **New Features**
  * Audio status indicator in Settings: when previewing alert sounds, a warning hint with a speaker icon appears if the system output is muted or volume is very low.
  * Preview controls now reflect alert-sound enablement and show the hint only while previewing or when alerts are enabled/disabled.

### DIFF
--- a/Sources/SettingsView.swift
+++ b/Sources/SettingsView.swift
@@ -321,6 +321,7 @@ struct GeneralSettingsView: View {
     @State private var keyValidationSuccess = false
     @State private var customVocabularyInput: String = ""
     @State private var micPermissionGranted = false
+    @State private var showMutedHint = false
     @StateObject private var githubCache = GitHubMetadataCache.shared
     @ObservedObject private var updateManager = UpdateManager.shared
     private let freeflowRepoURL = URL(string: "https://github.com/zachlatta/freeflow")!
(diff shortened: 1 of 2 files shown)


ojhurst:
Good catch on the onChange edge case — fixed in bb2c023. The hint now clears when alert sounds are toggled off.

marcbodea:
LGTM

PR #118, opened 2026-04-22
Add pulsating activity to waveform view

marcbodea:
## Summary
- Added an optional activity pulse mode to `WaveformView` so the waveform can animate while recording is active.
- Used `TimelineView` to drive a subtle traveling-wave/shimmer effect across the bars at 30 fps.
- Kept the original static waveform behavior when the pulse is disabled, and enabled it only during the recording phase.

## Testing
- Not run
- Verified the implementation diff updates only `Sources/RecordingOverlay.swift`
- Checked that the waveform still renders in the existing `WaveformView` layout with the same frame height

## Summary by CodeRabbit

* **New Features**
  * Waveform display now features an animated activity pulse effect during recording, with traveling wave and shimmer animations for enhanced visual feedback. The pulse automatically activates when recording begins and deactivates during other states.

* **Improvements**
  * Optimized waveform rendering to support smooth time-based animations at approximately 30 frames per second.

### DIFF
--- a/Sources/RecordingOverlay.swift
+++ b/Sources/RecordingOverlay.swift
@@ -305,15 +305,29 @@ struct WaveformBar: View {
 
 struct WaveformView: View {
     let audioLevel: Float
+    var showsActivityPulse = false
 
     private static let barCount = 9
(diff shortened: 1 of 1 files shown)


marcbodea:
@coderabbitai review

PR #141, opened 2026-04-24
Fix paste failing on Colemak-DH and other non-QWERTY layouts

cathreya:
## Summary

- `pasteAtCursor()` hardcoded `virtualKey: 9` (the physical 'V' key on QWERTY). On layouts that relocate V — notably Colemak-DH — that physical key produces a different character, so `Cmd+keycode-9` is not interpreted as `Cmd+V` by the OS and the transcribed text never pastes.
- Fix by dynamically looking up the physical key that produces `'v'` in the current keyboard input source via `TISCopyCurrentKeyboardInputSource` + `UCKeyTranslate`, with a fallback to keycode 9 if the lookup fails.
- Standard Colemak keeps V at the QWERTY position so it was unaffected; this primarily fixes Colemak-DH users (and any other layout that moves V).

## Test plan

- [x] Builds cleanly (`make CODESIGN_IDENTITY=- run`)
- [x] Verified transcribed text pastes correctly on Colemak in multiple apps (Terminal, browser text fields, Notes)
- [ ] (Recommended) Verify on standard QWERTY — should be unchanged since the lookup returns keycode 9 there as well
- [ ] (Recommended) Verify on another non-QWERTY layout (Dvorak, Colemak-DH) if a reviewer has one available

## Bug Fixes

* Improved paste functionality to dynamically detect and use the correct keyboard key codes based on your current keyboard layout, ensuring paste operations work reliably with non-QWERTY and other custom keyboard configurations.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -5,6 +5,7 @@ import AVFoundation
 import ServiceManagement
 import ApplicationServices
 import ScreenCaptureKit
+import Carbon
 import os.log
 private let recordingLog = OSLog(subsystem: "com.zachlatta.freeflow", category: "Recording")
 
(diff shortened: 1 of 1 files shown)


marcbodea:
will merge this but I think we should work on a better way to do paste, some ideas from codex:

1. Try Accessibility direct insertion first
Use AXUIElementCopySystemWideElement(), get kAXFocusedUIElementAttribute, then try setting text/range attributes such as kAXSelectedTextAttribute or kAXValueAttribute.
This is cleaner when it works because it avoids keyboard events and clipboard mutation. But it will fail or behave inconsistently in many apps, especially Electron, browsers, terminals, custom editors, password fields, and rich text controls.

2. Fall back to app-native paste
Keep the current pasteboard flow, but instead of hardcoding Cmd-V as the only mechanism, possibly try invoking the focused app’s Edit > Paste menu item through Accessibility. That avoids keyboard layout/shortcut issues, but menu traversal is slower and localized menu titles make it less robust unless handled carefully.

3. Fall back to synthetic Cmd-V
current approach

PR #146, opened 2026-04-25
Add stamped waveform menu bar icon for FreeFlow Dev

ojhurst:
## Summary

When developing FreeFlow alongside the prod app, both bundles' menu bar icons look identical — a small waveform SF Symbol — so a developer with both installed can't tell which build's menu bar entry they're looking at. This PR adds a stamped/inverse-waveform variant for the dev bundle: a filled rounded rectangle with the waveform cut out as negative space (Drive-style treatment), rendered as a template image so the menu bar tints it to whatever color the label uses. Only swaps the idle-state icon; recording and transcribing keep their existing SF Symbols so the state signal isn't lost.

## Why

After #137 added a distinct app icon for dev builds (the hammer overlay) and #136 made `CFBundleName` honest, the one place a developer still couldn't tell prod from dev at a glance was the menu bar. The dock icon, the app switcher, the Finder thumbnail — all distinguished. The menu bar was the holdout, and that's the surface a dictation-app user looks at most often during development.

The negative-space "stamp" treatment is intentionally distinct from the prod icon (which is the live waveform symbol). Prod stays familiar, dev reads at a glance as "this is the developer build."

## What it does

- New private helper `MenuBarLabel.devIdleIcon` constructs an `NSImage` at runtime: filled rounded rectangle, then 9 vertical bar paths cut out via `NSBezierPath` + even-odd fill, marked as a template image (`isTemplate = true`) so AppKit tints it to match the menu bar appearance.
- `MenuBarLabel.body` now picks the icon based on state AND bundle:
  - `isRecording` → `record.circle` SF Symbol (unchanged)
  - `isTranscribing` → `ellipsis.circle` SF Symbol (unchanged)
  - idle, dev bundle (`CFBundleName == "FreeFlow Dev"`) → the new stamped image
  - idle, prod → `waveform` SF Symbol (unchanged)
- No new resources committed — the icon is drawn at runtime so there's no `.png` or `.pdf` to ship.

## Why gate on CFBundleName, not a build flag

The Makefile pattern in this repo (and in #136, #137, the upcoming PRs in this series) keys behavior off `CFBundleName` because that's the value `$(APP_NAME)` writes into Info.plist at build time. Same single source of truth, no compile-time flags to keep in sync.

## Test plan

- [x] Builds cleanly with `make` on macOS 13+ arm64
- [x] Prod bundle (`APP_NAME=FreeFlow`) — menu bar shows the `waveform` SF Symbol when idle, identical to before
- [x] Dev bundle (`APP_NAME=FreeFlow Dev`) — menu bar shows the new stamped/inverse-waveform image when idle; tints correctly under both light and dark menu bars (template image)
- [x] Recording state shows `record.circle` regardless of bundle (state signal preserved)
- [x] Transcribing state shows `ellipsis.circle` regardless of bundle (state signal preserved)
- [x] No new files added to `Resources/`, no new compile-time dependencies

## Summary by CodeRabbit

* **New Features**
  * Menu bar icon now dynamically updates based on recording and transcribing states.

* **Refactor**
  * Improved menu bar icon display logic with enhanced state-aware rendering for development builds.


marcbodea:
just a heads up that you committed .claude/dev-log

Issue #109, opened 2026-04-21
Show a "taking longer than normal" hint when transcription stalls


ojhurst:
Small UX suggestion around the waiting state.

![The FreeFlow recording pill showing the waveform dots](https://assets.themarketingshow.com/bugs/freeflow/waveform-pill-2026-04-21.png)

Sometimes after I stop speaking there is a pause before the text lands. On a normal trip it is quick. On a slow trip (network hiccup, whatever) I am left wondering whether FreeFlow stalled out or is still working. Wispr Flow does a nice thing here: after a certain threshold it shows a small "taking a little longer than normal" hint so you know it is still going.

Would love that same hint in FreeFlow. After N seconds with no response from the API, the indicator changes state (pulse, or a small caption under the audio waveform) to signal "still working, just slow."

## Bonus idea

Would not mind if the transcribed words showed up right where the waveform is drawn. That area is already in my eyeline since I am staring at it while I talk, so seeing the text land there would be a nice extra confirmation.

---

**Environment**
- FreeFlow: 0.3.0 — `build-20260421-154112-1167337` (read from `CFBundleShortVersionString` + `FreeFlowBuildTag` in Info.plist)
- macOS: 26.4.1 (arm64)

marcbodea:
closed by #151

PR #151, opened 2026-04-29
Show transcribing spinner immediately with delayed fallback

marcbodea:
## Summary
- Show the transcribing indicator immediately instead of waiting for a short delay.
- Replace the old waveform-only processing state with a combined indicator that starts as an animated waveform and falls back to a spinner after 1 second.
- Simplify `AppState` by removing the separate transcribing indicator task and related delay bookkeeping.

## Testing
- Not run
- Reviewed the overlay state flow to confirm transcribing now enters the indicator immediately.
- Reviewed the new `ProcessingIndicatorView` transition from waveform to spinner after the timeout.

Screen Recording 2026-04-29 at 17 00 49

## Summary by CodeRabbit

* **New Features**
  * Processing indicator now starts as a compact 5-bar capsule waveform and, after ~1s, transitions into an animated rotating spinner.
  * Transcribing UI now appears immediately when transcription begins; recording visuals are shown only during active recording.

* **Style**
  * Refreshed waveform visuals and smoother animation/transitions for recording and transcribing states.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -2319,7 +2319,7 @@ final class AppState: ObservableObject, @unchecked Sendable {
         statusText = "Preparing audio..."
         errorMessage = nil
         playAlertSound(named: "Pop")
-        overlayManager.prepareForTranscribing()
+        overlayManager.showTranscribing()
         audioRecorder.stopRecording { [weak self] fileURL in
             guard let self else { return }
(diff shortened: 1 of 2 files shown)


marcbodea:
kind of closes #109? what do you think @ojhurst

ojhurst:
yes it would

PR #168, opened 2026-04-30
Fix Advanced Provider Settings click race + grey out unused Reset buttons

ojhurst:
## Why

Two small fixes in the API Key card area.

**Click race on the disclosure.** The "Advanced Provider Settings" disclosure used both `DisclosureGroup`'s built-in label tap AND a manual `.onTapGesture` on the same label. Clicks on the label often fired both handlers in the same frame, the toggles cancelled, and the row "did not take." Replaced with a manual `VStack` + `Button` + chevron pattern — one click target, full-row hit area, animated expand/collapse.

FreeFlow Settings panel showing the API Key card with the Advanced Provider Settings disclosure row in its new collapsed state. The green status line and key field above it are blurred to indicate they are unrelated to this PR.

*Settings view with the new disclosure row in focus (chevron + label, single click). The blurred rows above are unrelated to this PR.*

**Reset buttons that did nothing.** The five "Reset to Default" buttons in `ProviderSettingsFields` were always enabled and always clickable, even when the field already held the default value. Now they disable when there is nothing to reset.

Advanced Provider Settings panel with all Reset to Default buttons greyed out because the adjacent fields hold default values

*All five Reset to Default buttons greyed out — every adjacent field holds the default value, so there is nothing to reset.*

## Files

- `Sources/SettingsView.swift`

## Test plan

- [x] Builds cleanly with `make`
- [ ] Click anywhere on the "Advanced Provider Settings" row → expands once, no double-toggle
- [ ] All five "Reset to Default" buttons grey out when the adjacent field holds the default value, enable when the user types something different

### DIFF
--- a/Sources/SettingsView.swift
+++ b/Sources/SettingsView.swift
@@ -143,6 +143,7 @@ struct ProviderSettingsFields: View {
                     appState.apiBaseURL = AppState.defaultAPIBaseURL
                 }
                 .font(.caption)
+                .disabled(apiBaseURLInput == AppState.defaultAPIBaseURL)
             }
 
             if showsModelDescription {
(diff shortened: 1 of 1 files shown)


marcbodea:
Disabling the reset to default buttons when they don't do anything makes sense.

I haven't had the issue with clicks cancelling each other out when opening the advanced settings menu, and the animation in the current build is a bit wonky with the advanced settings popping up on top of the other settings
Screen Recording 2026-05-01 at 11 05 54

PR #170, opened 2026-04-30
Masked API key display in Settings and Setup wizard with click-to-edit

ojhurst:
## Why

The API Key field used to render the saved key as a featureless row of dots, indistinguishable from any other masked password field. Once you had pasted a key, you could not tell which one was loaded. Months later when swapping providers, the only signal was "this field has dots in it" — not which key, not which provider.

This shows the saved key as `prefix6 + 16 dots + suffix6` (`gsk_5B••••••••••••••••zX9p2c`) when the field is not focused. Tapping the field swaps to a regular `TextField` showing the same content with asterisks in place of the dots, so Cmd+A + paste replaces the value cleanly. On blur, the field returns to the dotted display.

Same treatment lands in the Setup wizard so re-running setup with a saved key shows the readable display rather than an opaque row of dots.

## Stacked on #169

This PR depends on #169 (provider feedback). The cumulative diff against `main` includes both PRs' changes. Once #169 merges, this PR will rebase to show only the masked-key surface (~138 lines).

## What changes

- **`Sources/SettingsView.swift`** — `apiKeySection` swaps the `SecureField` for a `ZStack` of an editable `TextField` plus a masked `Text` overlay. New helpers: `maskedDisplayKey(_:char:)`, `editableMaskedKeyBinding`, `shouldRevealMaskedKey`. New `@FocusState apiKeyFocused`.
- **`Sources/SetupView.swift`** — same `ZStack` pattern in the wizard's API Key step. Same three helpers.

## Test plan

- [x] Builds cleanly with `make`
- [ ] Save a key in Settings → blur the field → field shows `prefix6 + 16 dots + suffix6`
- [ ] Click the masked display → field swaps to editable TextField with `prefix6 + 16 asterisks + suffix6`
- [ ] Cmd+A + paste a fresh key → value replaces cleanly, no asterisk leakage
- [ ] Blur the field → returns to dotted display with the new key
- [ ] Re-run setup → API Key step shows the same masked display for a saved key

## Summary by CodeRabbit

* **New Features**
  * API key input fields now automatically mask displayed values when unfocused, showing only beginning and ending characters for enhanced security
  * Added automatic API provider detection with alerts when the provider doesn't match your configured host
  * Consolidated API key validation status into a single indicator banner for clearer feedback

PR #171, opened 2026-04-30
Confirm before switching Base URL on detected provider mismatch

ojhurst:
## Why

Pasting an API key from a different provider than the one the Base URL points at used to either silently fail with a 401 (because the old Base URL got the wrong-provider key) or, if a silent auto-switch was wired up, flip URL+models to whatever the key prefix detected — both paths take agency away from the user.

This adds a verify-first auto-switch with explicit confirmation:

1. Field blur or 400ms after paste fires `autoSwitchProviderIfNeeded`.
2. If the typed key's provider does not match the configured Base URL host's provider AND there is a preset for the typed key's provider, the key is validated against THAT preset's URL.
3. On success, the UI surfaces a blue banner with a `Switch to OpenAI` button on the right. The user explicitly opts in.
4. Button click applies the preset (URL + four model fields) and lights the green success label.
5. On validation failure, a specific error replaces the banner — current settings are preserved.

API Key card showing a blue checkmark banner that reads OpenAI key validated, Switch Base URL from Groq to OpenAI question mark, with a Switch to OpenAI button on the right

*The blue confirmation banner. User pasted an OpenAI key while the Base URL still pointed at Groq. The app validated the key against `api.openai.com` (preserving the user's existing Groq settings until they opt in), and surfaced the Switch button.*

Bidirectional: works in both directions (Groq → OpenAI, OpenAI → Groq) via the same provider-preset lookup. Pending state clears when the user types in the field.

## Stacked on #169 + #170

This PR builds on the provider-feedback PR (#169, adds `detectProvider` helpers) and the masked-key PR (#170, adds the `apiKeyFocused` state used for the blur trigger). The cumulative diff against `main` includes all three. Once the parents merge, this PR rebases to show only the confirm-before-switch surface (~174 lines).

## Surface

- `Sources/TranscriptionService.swift` — `validateAPIKeyDetailed` (returns `nil` on success or a specific error message), `ProviderPreset` struct, `providerPreset(forName:)` lookup table for Groq + OpenAI.
- `Sources/SettingsView.swift` — `PendingProviderSwitch` state, `autoSwitchProviderIfNeeded`, `applyPendingProviderSwitch`, debounced task scheduling on `apiKeyInput.onChange`, blur trigger on `apiKeyFocused.onChange`, banner UI in `apiKeySection`.

## Test plan

- [x] Builds cleanly with `make`
- [ ] With Groq configured, paste a valid OpenAI key → blue banner appears within 500ms with a `Switch to OpenAI` button
- [ ] Click the Switch button → Base URL + four model fields swap to OpenAI defaults, green success label reads `API key saved — Validated as OpenAI`
- [ ] With Groq configured, paste a fake `sk-` key → red error banner reads `OpenAI rejected the key (...). Kept your Groq settings.` Base URL stays Groq.
- [ ] Type a partial key (still has prefix `sk-`) → no API call yet, debounce holds
- [ ] Same flow in reverse (OpenAI configured, paste a `gsk_` key) works

## Release Notes

* **New Features**
  * Enhanced API key security with masked display and focus-aware preview modes
  * Real-time validation with detailed error messages for configuration issues
  * Automatic provider detection from API key and base URL configuration
  * Visual status indicators showing validation results and pending provider switches

### DIFF
--- a/Sources/SettingsView.swift
+++ b/Sources/SettingsView.swift
@@ -460,7 +460,15 @@ struct GeneralSettingsView: View {
     @State private var keyValidationError: String?
     @State private var keyValidationSuccess = false
     @State private var keyValidationProviderLabel: String?
+    @State private var pendingProviderSwitch: PendingProviderSwitch?
+    @State private var pendingValidationTask: Task<Void, Never>?
     @FocusState private var apiKeyFocused: Bool
+
+    private struct PendingProviderSwitch {
+        let preset: TranscriptionService.ProviderPreset
+        let fromProviderName: String
+        let key: String
+    }
     @State private var customVocabularyInput: String = ""
     @State private var micPermissionGranted = false
     @State private var showMutedHint = false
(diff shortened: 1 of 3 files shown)


marcbodea:
In general I think this and #169 add a lot of code that helps a small number of more "advanced" users which should be able to understand the errors they are getting and how to fix them.

I am not sure increasing the surface area and potential for bugs to help a small subset of users is a good trade-off.

PR #183, opened 2026-05-04
[codex] Add Caps Lock shortcut preset

richstokes:
## Summary
- Adds Caps Lock as a built-in shortcut preset alongside Fn, Right Option, and F5.
- Wires Caps Lock key code 57 through shortcut display, preset bindings, modifier matching, and local/global shortcut capture.
- Tracks Caps Lock from flagsChanged state explicitly so regular modifier snapshots do not infer it from the system latched flag alone.

## Validation
- `git diff --check`
- `make CODESIGN_IDENTITY=-`


richstokes:
Closing this one and replacing it with the same change from a branch named caps-lock-shortcut.

PR #184, opened 2026-05-04
Add Caps Lock shortcut preset

richstokes:
## Summary
- Adds Caps Lock as a built-in shortcut preset alongside Fn, Right Option, and F5.
- Wires Caps Lock key code 57 through shortcut display, preset bindings, modifier matching, and local/global shortcut capture.
- Tracks Caps Lock physical key state separately from the macOS latched Caps Lock flag so hold mode can stop on release.

## Summary by CodeRabbit

* **New Features**
  * Caps Lock is now a first-class shortcut modifier with visible labels, an ordered display name, and a preset.
  * Optional physical Caps Lock monitoring can be enabled for more reliable detection.

* **Bug Fixes**
  * Improved Caps Lock state accuracy across modifier events and lifecycle (monitor stops on teardown/reset).
  * Physical monitor setup failures now surface explicit error reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/zachlatta/freeflow/pull/184?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/CapsLockPhysicalKeyMonitor.swift
+++ b/Sources/CapsLockPhysicalKeyMonitor.swift
@@ -18,6 +18,7 @@ enum CapsLockPhysicalKeyMonitorError: LocalizedError {
 final class CapsLockPhysicalKeyMonitor {
     private var manager: IOHIDManager?
     private let runLoopMode = CFRunLoopMode.commonModes.rawValue
+    private var capsLockIsDown = false
 
     var onStateChanged: ((Bool) -> Void)?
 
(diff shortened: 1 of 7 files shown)


marcbodea:
It seems that using caps lock as the hold shortcut makes it function more or less as a tap shortcut. So the dictation does not stop after I release caps lock but rather continues until I press it again. Could you check and let me know if it behaves similarly for you?

richstokes:
@marcbodea Yep, the issue was that Caps Lock was being read from the macOS modifier flags, which reflect the latched on/off state rather than the physical key being held. I pushed a fix that tracks Caps Lock physical down/up via HID for Caps Lock shortcuts, so hold mode can stop on release instead of waiting for the next press.

marcbodea:
Hi, I am having trouble getting it to work on my machine, added some additional logging and it seems that the physical monitor is not receiving a signal, but the normal event tap does.

I would be much more comfortable merging a PR that adds caps lock just as a tap shortcut, since I want to be cautious adding a feature that requires a separate physical monitoring path.

I am open to redoing the shortcut monitor overall if we can find a better, more flexible and reliable solution.

PR #188, opened 2026-05-06
Add mouse-button and chord dictation shortcuts

alexbreen7:
## Why this matters

Many people already have every easy keyboard combo claimed by the OS or another app. **Mouse buttons** (especially middle click and side buttons) are often unused and sit under the same hand you use for the trackpad or mouse, which makes them a natural **push-to-talk** or **tap-to-toggle** surface.

**Chords** (e.g. **middle-click + Space**) add a deliberate second step so dictation does not start from a single accidental click. The same plumbing also supports **two-key** keyboard chords when that fits a layout better.

**Left click is intentionally unsupported** as a shortcut or chord partner—it is too easy to fire by accident while editing text or clicking around the UI.

## High-level summary

- Record and use **mouse-based shortcuts** alongside existing keyboard shortcuts (hold + toggle).
- Record **two-part shortcuts**: **mouse button + one non-modifier key**, or **two non-modifier keys**, with modifiers handled the same way as before.
- Keep **global shortcut behavior** consistent: mouse down/up participates in the matcher and can be **consumed** when it matches an active binding.

## Technical approach

- **`ShortcutBinding`** gains optional **`chordKeyCode`** / **`chordMouseButton`**, a new **`ShortcutBindingKind.mouseButton`**, and **`id` / `displayName` / specificityScore`** updates so chords stay distinct and readable in the UI.
- **`ShortcutInputEvent.mouseChanged(button:isDown:)`** plus **`ShortcutInputState.pressedMouseButtons`** feed the matcher; **`bindingIsActive`** for `.key` / `.mouseButton` requires both primary and chord inputs when configured.
- **`GlobalShortcutBackend`** extends the CGEvent tap to **mouse down/up** (left/right/other) and forwards them through the same **`onInputEvent`** path as keys.
- **`LocalShortcutCaptureBackend`** tracks **currently pressed non-modifier keys and mouse buttons**, listens for **key-up** and **mouse up/down**, so capture can snapshot a chord without treating it as unrelated key events only.
- **`ShortcutComponents` / capture UI** updates the prompt and wires mouse/key-up handlers to **`ShortcutBinding.fromCaptureState`**; completion still uses Enter / Done as before.
- **`AppState`** adds validation copy when someone tries to bind **left mouse button** as primary or chord partner.

## Summary by CodeRabbit

* **New Features**
  * Add custom mouse-button shortcuts, including chords that combine non-primary mouse buttons with keys.
  * Capture UI updated with clearer instructions and improved mouse/key capture behavior.

* **Bug Fixes**
  * Validation now rejects unsupported left-button-only shortcuts and guides users to side/middle/right buttons.
  * Input handling correctly recognizes and respects mouse-button bindings (including chorded bindings).

[Review Change Stack](https://app.coderabbit.ai/change-stack/zachlatta/freeflow/pull/188?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


kristian-nst:
Yes pls! I'm using Button5 (sidebuttons) for Flow, but I couldn't set it to that in Freeflow :/

marcbodea:
@coderabbitai full review

marcbodea:
@alexbreen7 can you take a look at the comments coderabbit posted? If valid please fix otherwise reply to it so it knows why it's not needed.

Thanks!

alexbreen7:
@marcbodea Thanks for the review — we addressed the feedback. Summary of what changed and how it differs from the suggestions:

Implemented
1. Symmetric keyUp / mouseUp suppression (LocalShortcutCaptureBackend)

keyUp now returns nil, matching keyDown.
mouseUp now returns nil for non-left buttons and passes through button 0, matching mouseDown.
This avoids unmatched release events during capture while still allowing Done / Cancel to receive left-click up events.

2. Stop release handlers from clobbering currentBinding (ShortcutComponents)

The review suggested guarding onKeyUpEvent / onMouseUpEvent so we only assign when fromCaptureState returns a valid binding and skip left-button mouse-up.

We took a slightly different approach: removed those handlers entirely and latch currentBinding only on down paths (onKeyDownEvent, onMouseDownEvent, and modifier updates via onInputEvent).

Why: The up handlers always ran against post-release state, so they tended to clear or downgrade a valid chord (e.g. before Done / Enter). That conflicted with the UI copy (“Press Esc or Enter to save” after releasing the combo). Latching on down matches that flow and is simpler than conditional assigns on every release.

3. Remove chordMouseButton (nit)

Removed the property and related matcher / validation code. Capture already encodes key+mouse combos as mouse-primary + chordKeyCode; nothing in the app ever set chordMouseButton.

Intentionally unchanged
onKeyUpEvent / onMouseUpEvent on LocalShortcutCaptureBackend

We left these properties and the internal onKeyUpEvent? / onMouseUpEvent? invocations in the backend. Nothing in the capture UI wires them up anymore, but the hooks remain if we need them later. Removing them would be cosmetic only.

Testing
Manually verified: record a chord → release → Enter or Done saves correctly; Esc cancels; Done / Cancel still work with left click during capture.

PR #189, opened 2026-05-06
Dictation: migrate legacy “audio interruption mode” to mute toggle storage

alexbreen7:
## What problem this solves

While you dictate, sound from the Mac’s **default output** (music, calls, notification chimes, etc.) can still play. That is distracting, can bleed into the microphone, and makes it harder to hear yourself think. FreeFlow can **temporarily mute that default output for the duration of a dictation session** and **return it to how it was afterward** (Settings → **Mute audio when dictation starts**).

This is **output mute** via Core Audio, not a global “media play/pause” control for third-party players.

## High-level summary

- **User-facing behavior:** Same as today on `main` when the toggle is on: mute default output for active dictation, restore prior mute state when dictation ends.
- **What this PR changes:** Makes **saved preferences reliable** for installs that still have the older `dictation_audio_interruption_mode` string (from when the app stored a `mute` / `pause` / `off` style mode) by migrating once to the current boolean `dictation_audio_interruption_enabled`, then dropping the legacy key.

## Code details

- If `dictation_audio_interruption_enabled` is already present, its value is used as today.
- Otherwise, if `dictation_audio_interruption_mode` exists: `mute` → enabled, any other value → disabled; persist the bool and remove the mode key.
- When the user toggles the checkbox, also remove `dictation_audio_interruption_mode` so stale values cannot linger.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -229,7 +229,9 @@ final class AppState: ObservableObject, @unchecked Sendable {
     private let outputLanguageStorageKey = "output_language"
     private let realtimeStreamingEnabledStorageKey = "realtime_streaming_enabled"
     private let realtimeStreamingModelStorageKey = "realtime_streaming_model"
-    private let dictationAudioInterruptionEnabledStorageKey = "dictation_audio_interruption_enabled"
+    private static let dictationAudioInterruptionEnabledStorageKey = "dictation_audio_interruption_enabled"
+    /// Migrates users who stored the picker `mute`/`pause`/`disabled` variant.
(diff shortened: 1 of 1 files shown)


alexbreen7:
Closing context: I was working from a fork `main` that was quite a bit behind upstream, so it looked like this mute-during-dictation behavior wasn’t land yet. After syncing with current `origin/main`, it’s clear the feature is already there (settings toggle + output mute path), so this migration-only PR is redundant for the normal case and doesn’t add enough to justify merging.

Thanks for the patience — I’m fine to close this PR.

PR #191, opened 2026-05-07
Add Groq backup-key auto-rotation for free-tier survivability

ojhurst:
## What

Adds an optional Backup API Key field. When the active Groq key returns a 429 with `tokens per day` (the per-org daily cap), the post-processing path swaps to the other key, retries the failed request once, and surfaces the rotation in the menu-bar pill.

## Why

Groq's free tier exhausts quickly under heavy command-mode use — once the per-day token cap on the active org is hit, every subsequent post-processing request 429s for the rest of the day. Today the user sees dictation silently stop working with no rotation path. This PR lets users keep going by configuring a key from a second Groq account as a backup; rotation is automatic when the daily cap fires.

Per-minute throttling does NOT trigger rotation. The existing model fallback chain (primary model → fallback model on 429) already handles that case well.

## How

- `apiKeyBackup` and `activeKeyIsBackup` added to `AppState`.
- `activeAPIKey` computed property resolves to whichever key is live.
- `rotateAPIKeyIfDailyExhausted(error:)` inspects the error, flips the active key when the other slot has a value, fires a pill toast, and returns `true` so the caller can retry.
- Both `processTranscript` catch blocks (Edit Mode and post-processing) invoke rotation and retry once with a freshly-constructed `PostProcessingService` so the new key takes effect mid-flight.
- `SettingsView` gains a "Backup API Key" card directly below the primary. Active-key indicator and manual "Switch back to primary" button surface only when the backup is the live key.

## UI

(Screenshot is from a downstream variant that uses a masked-with-peek field for both keys; the upstream version in this PR uses `SecureField` matching the existing primary API Key pattern. Structural placement and the active-key indicator are identical.)

## Behaviour notes

- Active-key state persists across launches — an exhausted key is not retried after restart.
- Empty backup field disables rotation entirely; behaviour is unchanged for users who do not configure one.
- Same Groq account on both slots provides no rotation benefit (daily caps are per-org, not per-key); the documentation copy notes this implicitly by saying "leave blank to disable rotation."

## Dependency

Depends on #190 (in-pill error notifications) for the `showError(_:)` surfacing mechanism. Once #190 lands, this PR rebases onto main and the calls resolve. Code is presented assuming `showError` exists.

## Release Notes

* **New Features**
  * Configure backup API key in settings for failover protection
  * Automatic key rotation when daily usage limit is reached
  * Manual switching between primary and backup API keys available
  * Visual indicator displays currently active API key

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -198,6 +198,13 @@ final class AppState: ObservableObject, @unchecked Sendable {
     }
 
     private let apiKeyStorageKey = "groq_api_key"
+    // Backup Groq key for free-tier survivability — when the active key
+    // returns a 429 with `tokens per day` (per-org daily cap exhausted),
+    // the post-processing path swaps to the other key and retries once.
+    // Per-minute throttling does NOT trigger rotation; the existing model
+    // fallback handles that.
+    private let apiKeyBackupStorageKey = "groq_api_key_backup"
+    private let activeKeyIsBackupStorageKey = "groq_active_key_is_backup"
     private let apiBaseURLStorageKey = "api_base_url"
     private let transcriptionModelStorageKey = "transcription_model"
     private let transcriptionAPIURLStorageKey = "transcription_api_url"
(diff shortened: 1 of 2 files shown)


marcbodea:
Would want to check groq terms before merging. Don't want to step on their toes by allowing/encouraging a breach since they already offer a generous free tier

ojhurst:
Fair point

ojhurst:
Even dynamically changing to another provider would be nice though.

marcbodea:
Groq’s Services Agreement says customers may not use the services “in a manner intended to avoid incurring Fees.” This PR explicitly rotates keys after daily cap exhaustion, so I am closing this to avoid any potential issues.

PR #192, opened 2026-05-07
Add multi-display picker for the recording overlay

ojhurst:
On multi-monitor setups, the recording overlay drops down on whichever screen has the active key window because the layout reads from NSScreen.main. That makes the pill follow focus across monitors, which is disorienting when the user wants it pinned to a specific display (e.g. always the laptop screen, or always the external).

This change adds a "Show on" picker to a new Overlay Display settings card with three options:

  - Active window (default) — current behavior, follows focus via NSScreen.main. Default for backward compatibility.
  - Primary display — always NSScreen.screens.first (the menu-bar display).
  - Each connected screen by its localizedName.

The chosen NSScreen displayID is stored in UserDefaults under `overlay_display_id`. If a saved display is unplugged, the resolver falls back to the primary. The picker live-updates when monitors are attached or removed via NSApplication.didChangeScreenParametersNotification, and scales to any number of connected displays (tested with two, designed for N).

The overlay's notch math (screenHasNotch, notchWidth, notchOverlap) now reads from the chosen screen too, so a notched MacBook paired with an external monitor renders correctly regardless of which display is selected.

Single-display users see no behavior change.

## Tested

On a Mac Studio with two external displays attached:

- Default (`overlay_display_id` unset / 0) — pill followed focus across both screens, matching pre-PR behavior. Backward compat verified.
- Active window — explicit selection, same follow-focus behavior.
- Primary display — pill stayed on the menu-bar display regardless of where the active window was.
- Specific primary display by localizedName — pinned correctly.
- Specific second display by localizedName — pinned correctly.

Overlay Display picker showing Active window default, Primary display, and connected screen names

## Summary by CodeRabbit

* **New Features**
  * Choose which physical display shows the recording overlay via a new "Overlay Display" setting (options: active window, primary display, or any connected display).
  * Overlay repositions, updates safe-area/notch padding, and matches hide/show animations to the selected display.
  * Display list refreshes automatically when monitors change and the selection is persisted across launches.

* **Documentation**
  * Added 1.0.0 release notes marking FreeFlow as stable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/zachlatta/freeflow/pull/192?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ojhurst:
Thanks for the review! Addressed finding 1 in 870ac60 — added `.accessibilityLabel("Show on")` to the Picker so VoiceOver reads the same label a sighted user sees on the row.

On finding 2 (clamping the bound `overlayDisplayID` when a display is unplugged): I'd like to leave this as-is. The runtime resolver in `RecordingOverlay.targetScreen` already falls back to the primary display when the saved displayID is no longer connected, so overlay placement stays correct. Resetting the binding on `didChangeScreenParametersNotification` would actively destroy the user's stored preference — when they replug the monitor, their choice is gone. The bot's own note flags the tension ("preserve user choice when possible"), and I think preserving the stored intent across hot-plug events outweighs the brief blank-picker state when the saved display is offline.

Happy to revisit if there's a better UX (e.g. rendering an "(offline)" entry for the saved-but-missing display so the picker doesn't render blank), but that's a bigger change than the bot suggested and I'd want to discuss before adding it.

PR #199, opened 2026-05-20
feat: mark dictated text as transient so clipboard managers skip it

pianoandglass:
## Problem

Every dictation currently writes the transcript to the system clipboard, which causes **clipboard managers** (Maccy, Raycast, Paste, Clipy, Flycut, etc.) to record each dictated phrase in their history. This creates noise: the user never manually copied that text — it was just an intermediate step in the paste-at-cursor flow.

## Solution

Declare standard **transient pasteboard types** alongside `.string` when writing to the clipboard. Clipboard managers that respect the [org.nspasteboard conventions](https://github.com/nicke5012/TransientPasteboardType) will automatically skip recording the entry.

The four types declared are:

| Type | Purpose |
|---|---|
| `org.nspasteboard.TransientType` | Standard transient marker |
| `org.nspasteboard.ConcealedType` | Hides from clipboard history |
| `org.nspasteboard.AutoGeneratedType` | Signals auto-generated content |
| `de.petermaurer.TransientPasteboardType` | Legacy compatibility |

`declareTypes(_:owner:)` is used instead of `clearContents()` + `setString()` to declare all types atomically. Empty string values are set for each marker type, as some clipboard managers check for data presence rather than just the declared type.

## Behavior

- ✅ Text **still pastes normally** via Cmd-V in all apps
- ✅ Clipboard managers that respect transient types **skip** the entry
- ✅ Clipboard managers that don't respect transient types are **unaffected** (they still record it, same as before)
- ✅ Clipboard preserve/restore (`preserveClipboard`) continues to work correctly

## Testing

Tested with Maccy and Raycast. Dictated text no longer appears in clipboard history after dictation.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced clipboard handling when copying transcripts so clipboard/history tools can more reliably detect and preserve copied transcripts.
  * Standard paste behavior (Cmd‑V) is unchanged for end users; copying remains seamless while clipboard managers gain improved ability to store and index transcript entries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/zachlatta/freeflow/pull/199?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


marcbodea:
great PR, thanks for this!

PR #207, opened 2026-05-21
Add a Dictionary of deterministic transcript corrections

ojhurst:
## Summary

Adds a Dictionary — a list of user-defined `heard → replacement` rules applied to the final transcript as the last step of the dictation pipeline, after the LLM post-processing pass. A correction is a guaranteed find-and-replace: it fires every time, deterministically, and still applies when post-processing is disabled or the API is unavailable.

This is the same dictionary pattern found in other dictation tools such as WhisperFlow.

## Why, and how it differs from Custom Vocabulary

Custom Vocabulary feeds the cleanup model a soft spelling hint. It is probabilistic — the model may or may not act on it, and it does nothing when post-processing is off.

The Dictionary is a different mechanism: a deterministic find-and-replace on the final text. Not a hint to a model — a literal substitution rule. When the transcription gets a specific phrase wrong the same way every time, the user names exactly what came out and what they meant, and the rule fixes it on every dictation, guaranteed. The two are complementary: vocabulary for soft spelling guidance, the Dictionary for exact, known corrections.

## Screenshot

The Dictionary tab in Settings, showing the Corrections card with two example heard-to-replacement rules

## What's in it

- `Correction.swift` — a `Correction` model (heard, replacement, wholeWord, caseSensitive, enabled) and `DictionaryEngine`, a pure, dependency-free apply function. Whole-word matching uses a Unicode-aware boundary so a short rule like `cat → dog` does not rewrite `category`.
- `AppState` — a persisted `corrections` array and the pipeline hook: `DictionaryEngine.apply` runs on the final transcript inside `processTranscript`, on both the post-processing-success and raw-fallback paths.
- Settings — a new Dictionary tab with full add / edit / delete, a per-rule enable toggle, and a live preview in the editor.

No existing behavior changes when the corrections list is empty.

## Testing

Built and run first as a daily-driver feature in a personal fork, then re-implemented cleanly against v1.0.0 and validated again in a fresh build of that release as a feature patch.

Exercised by live dictation:
- Empty corrections list — pipeline unchanged, transcript passes through.
- Single correction, and multiple corrections within one utterance.
- Repeatability — the same utterance dictated repeatedly gives the same result every time.
- An arbitrary mapping (a phrase → an unrelated phrase) to confirm the transformation is the deterministic engine, not the cleanup model.
- Full add / edit / delete, including running with rules present, deleted, and re-added.
- The engine call site was temporarily instrumented to confirm it runs on every dictation with the expected rule count, then the instrumentation was removed once verified.

## Summary by CodeRabbit

* **New Features**
  * Added a Dictionary settings tab for managing custom transcript corrections.
  * Users can create, edit, and delete corrections that automatically apply to transcriptions.
  * Supports whole-word and case-sensitive matching options with a live preview when editing.
  * Corrections are persistently saved and can be individually enabled or disabled.

[Review Change Stack](https://app.coderabbit.ai/change-stack/zachlatta/freeflow/pull/207?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -24,6 +24,7 @@ enum SettingsTab: String, CaseIterable, Identifiable {
     case general
     case prompts
     case macros
+    case dictionary
     case runLog
     case debug
 
@@ -40,6 +41,7 @@ enum SettingsTab: String, CaseIterable, Identifiable {
         case .general: return "General"
         case .prompts: return "Prompts"
         case .macros: return "Voice Macros"
+        case .dictionary: return "Dictionary"
         case .runLog: return "Run Log"
         case .debug: return "Debug"
         }
@@ -50,6 +52,7 @@ enum SettingsTab: String, CaseIterable, Identifiable {
         case .general: return "gearshape"
         case .prompts: return "text.bubble"
         case .macros: return "music.mic"
+        case .dictionary: return "character.book.closed"
         case .runLog: return "clock.arrow.circlepath"
         case .debug: return "wrench.and.screwdriver"
         }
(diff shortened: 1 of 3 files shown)


marcbodea:
@ojhurst can you please take care of the coderabbit comments here and in the other open PRs? If they do not apply or are nonsense (it does happen), just reply directly to the comment and it will be resolved. Thanks!

PR #208, opened 2026-05-24
Add opt-out for sending screenshots to the context model

ojhurst:
## Summary

Adds a **"Use screen context for smarter dictation"** toggle in Settings → Prompts → Context Prompt, plus a **Skip** button on the Setup wizard's Screen Recording step. Both default to **on**, so existing users see zero behavior change.

## Why

The context inference is genuinely useful, but it requires the Screen Recording permission — and every dictation captures a JPEG of the active window and ships it to whatever inference endpoint the user has configured. There is currently no way to disable just the screenshot without revoking the OS permission entirely (which also kills the grant, so flipping it back on means re-granting through System Settings).

This PR adds a clean opt-out:

- Existing user who is happy with how things work: do nothing, defaults are unchanged.
- Privacy-conscious user who wants the screenshot off: one toggle in Settings.
- New user who does not want to grant the permission at all: Skip the step in Setup.

## How it works

`AppContextService.captureActiveWindowScreenshot` gates on a new `use_screenshot_context` UserDefaults flag (default `true`). When off, it returns the same `(nil, nil, error)` tuple shape as the existing permission-denied path — so the existing `screenshotDataURL == nil` fallback throughout the codebase keeps working without modification. No new code paths, just a clean early return with a user-facing error string that points back at the toggle.

The Run Log's "Capture Context" step (Step 1) gets a small no-image caption when `contextScreenshotDataURL` is nil but `contextScreenshotStatus` is non-empty — so the user can see at a glance why a given dictation has no screenshot attached.

Test button in the Context Prompt section refuses to run when the toggle is off, with a clear message pointing back at the toggle, instead of silently running a degraded test.

## Observability

Adds an `os_log` subsystem `com.zachlatta.freeflow` category `Context` that emits one line per capture outcome:

| Outcome | Log line |
|---|---|
| Toggle off | `screenshot skipped — disabled in settings (Use screen context for smarter dictation = off)` |
| Permission denied | `screenshot blocked — Screen Recording permission not granted` |
| Captured (active window) | `screenshot captured via active-window match (N bytes)` |
| Captured (focused title match) | `screenshot captured via focused-title match (N bytes)` |
| Captured (full-screen fallback) | `screenshot captured via full-screen fallback (N bytes)` |
| CGWindowListCreateImage nil | `screenshot failed — CGWindowListCreateImage returned nil` |
| Could not encode | `screenshot failed — image could not be encoded within size limits` |

Useful for verifying gate behavior from outside the app:

```
log show --predicate 'subsystem == "com.zachlatta.freeflow" AND category == "Context"' --info --last 10m --style compact
```

## What changed

- `Sources/AppState.swift` — new `useScreenshotContext` `@Published` Bool, persisted at UserDefaults key `use_screenshot_context`, default `true`.
- `Sources/AppContextService.swift` — `import os.log`, OSLog declaration, gate in `captureActiveWindowScreenshot`, and one `os_log` line on every outcome path.
- `Sources/SettingsView.swift` — adds the toggle subcard at the top of `contextPromptSection`, the test-button refusal, and the no-image caption in the Run Log's Capture Context step.
- `Sources/SetupView.swift` — adds a Skip button to the Screen Recording step that sets `useScreenshotContext = false` and advances without granting the permission.

**+79 / -0.** Purely additive, no upstream logic modified.

## Testing

Verified locally on a downstream build:

- **Default-on behavior is unchanged.** `log show --predicate 'subsystem == "com.zachlatta.freeflow" AND category == "Context"'` shows `screenshot captured via active-window match (N bytes)` per dictation, same as `main` today.
- **Toggle off skips the capture.** Same log shows `screenshot skipped — disabled in settings` instead, dictation still completes using the existing fallback context (app + window + selection), and the Run Log's Capture Context step displays the new "Screen context disabled in settings..." caption in place of the missing image.
- **Test button refuses with friendly error when the toggle is off.** Clicking Test Context Prompt with the toggle off renders `Screen context is disabled. Toggle "Use screen context for smarter dictation" on above to run this test.` instead of running a degraded test.

## Screenshots

The Settings toggle in Prompts → Context Prompt:

Use screen context for smarter dictation toggle in Settings

The Run Log's Capture Context step when the toggle is off (new no-image caption in this PR):

Run Log shows Screen context disabled in settings caption

The Test Context Prompt button refusal when the toggle is off:

Test Context Prompt refuses when screen context is disabled

## Notes

Default-on means zero regression for existing users. Defaults can flip later in a separate PR if that becomes desired.

## Release Notes

* **New Features**
  * Added toggle in settings to enable or disable screen context functionality
  * Added Skip option in setup flow for screen recording permissions
  * Added fallback status indicator when context capture is unavailable

* **Improvements**
  * Settings now persist across app sessions
  * Enhanced explanatory text for context capture behavior in settings

[Review Change Stack](https://app.coderabbit.ai/change-stack/zachlatta/freeflow/pull/208?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/AppContextService.swift
+++ b/Sources/AppContextService.swift
@@ -1,6 +1,7 @@
 import Foundation
 import ApplicationServices
 import AppKit
+import os.log
 
 struct AppSelectionSnapshot {
     let appName: String?
@@ -36,6 +37,7 @@ Return only two sentences, no labels, no markdown, no extra commentary.
 """
     static let defaultContextPromptDate = "2026-02-24"
     static let defaultScreenshotMaxDimension: CGFloat = 1024
+    fileprivate static let contextLog = OSLog(subsystem: "com.zachlatta.freeflow", category: "Context")
 
     private let apiKey: String
     private let baseURL: String
@@ -409,7 +411,19 @@ Selected text: \(selectedText ?? "None")
         appElement: AXUIElement,
         focusedWindowTitle: String?
     ) -> (dataURL: String?, mimeType: String?, error: String?) {
+        let useScreenshotContext = UserDefaults.standard.object(forKey: "use_screenshot_context") == nil
+            ? true
+            : UserDefaults.standard.bool(forKey: "use_screenshot_context")
+        if !useScreenshotContext {
+            os_log(.info, log: Self.contextLog, "screenshot skipped — disabled in settings (Use screen context for smarter dictation = off)")
+            return (
+                nil,
+                nil,
+                "Screen context disabled in settings. Toggle it on in Settings → Prompts → Context Prompt to use screenshots."
+            )
+        }
         if !CGPreflightScreenCaptureAccess() {
+            os_log(.info, log: Self.contextLog, "screenshot blocked — Screen Recording permission not granted")
             return (
                 nil,
                 nil,
@@ -485,6 +499,7 @@ Selected text: \(selectedText ?? "None")
                     compression: screenshotCompressionPrimary,
                     maxDimension: screenshotMaxDimension
                 ) {
+                    os_log(.info, log: Self.contextLog, "screenshot captured via active-window match (%{public}d bytes)", dataURL.count)
                     return (dataURL, "image/jpeg", nil)
                 }
             }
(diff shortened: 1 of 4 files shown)


marcbodea:
does not build

Sources/SettingsView.swift:1696:13: error: cannot find 'SettingsSubcard' in scope
1694 | 
1695 |         return VStack(alignment: .leading, spacing: 10) {
1696 |             SettingsSubcard {
     |             `- error: cannot find 'SettingsSubcard' in scope
1697 |                 VStack(alignment: .leading, spacing: 6) {
1698 |                     Toggle("Use screen context for smarter dictation", isOn: $appState.useScreenshotContext)
make: *** [build/FreeFlow Dev.app/Contents/MacOS/FreeFlow Dev] Error 1

PR #209, opened 2026-05-24
Add opt-out for sending app and window info to the context model

ojhurst:
## Summary

Adds a second toggle in Settings → Prompts → Context Prompt: **"Send app and window info"**. Pairs with the screenshot-context toggle in #208 so users can independently control both kinds of context the model sees — pixels (screenshot) and text strings (app name, window title, selected text). Default on, no behavior change for existing users.

## Why

Even with the screenshot toggle off, the context inference still sends the active app's name, bundle identifier, window title, and any selected text to the model. That string set alone is often enough to leak meaningful information — a browser window title contains the Gmail thread subject and the user's email address, a code editor's title bar reveals the active file, a SaaS dashboard's title gives away the account being viewed.

Some users would rather the model see none of that — accepting weaker context-aware cleanup in exchange for the strongest privacy posture. This toggle gives them that option, and pairs with #208 to make the privacy story complete:

| Screenshot toggle | App/window toggle | Model sees |
|---|---|---|
| ON | ON | screenshot + app + window + selection (today's behavior) |
| OFF | ON | app + window + selection only |
| ON | OFF | screenshot only (visual reading, no metadata) |
| OFF | OFF | nothing but the audio (fully blind dictation) |

## How it works

`AppContextService.collectContext` still performs the AX reads internally so the screenshot capture path can find the active window. But when the new `send_app_and_window_context` UserDefaults flag is off, `appName`, `bundleIdentifier`, `windowTitle`, and `selectedText` are scrubbed to `nil` before being passed to the LLM inference call OR stored on the returned `AppContext`. The downstream `pipelineHistory` and run-log display honor the `nil`s naturally.

When BOTH toggles are off, the LLM inference call is skipped entirely (no signal, no API call), and `currentActivity` is set to a stable string: `"You are dictating. App and window context are disabled in settings, and no screenshot was captured."` — so the run log makes the configuration explicit instead of silently producing nothing.

## Observability

One new `os_log` line under the existing `Context` category (added in #208):

```
app/window context skipped — disabled in settings (Send app and window info = off)
```

Pairs with the screenshot skip line. When both fire on the same dictation, the unified log shows two skip entries on the same thread at the same millisecond — the verifiable signature of fully blind dictation.

## What changed

- `Sources/AppState.swift` — new `sendAppAndWindowContext` `@Published` Bool, persisted at UserDefaults key `send_app_and_window_context`, default `true`.
- `Sources/AppContextService.swift` — adds the gate in `collectContext`; AX reads still happen internally (the screenshot path needs them), but the values are scrubbed before going to the LLM or to the returned `AppContext` when the toggle is off. Also short-circuits the LLM inference call when both toggles are off, since there is no signal worth spending an API call on.
- `Sources/SettingsView.swift` — adds the second toggle subcard right below the screenshot toggle.

**+50 / -6.** The -6 lines are a small refactor in `collectContext` to move the AX reads into local variables that get conditionally scrubbed.

## Testing

Verified locally:

- **Toggle off, screenshot on.** Unified log shows `app/window context skipped — disabled in settings` per dictation; screenshot still fires as expected. Inferred activity from the model becomes generic ("a coding environment", "an email application") instead of naming the specific app or thread.
- **Both toggles off.** Unified log shows BOTH skip lines on the same dictation, same thread, same millisecond. Run log's Capture Context step shows `"You are dictating. App and window context are disabled in settings, and no screenshot was captured."`

## Screenshots

The second toggle in Prompts → Context Prompt (sits directly below the screenshot toggle from #208):

Send app and window info toggle in Settings

## Notes

- Depends on #208 (this branch is stacked on `feat/optional-screenshot-context`). When #208 lands, this rebases to `main` cleanly. If #208 stalls, this stays a dependent and can be re-targeted.
- Default-on means zero regression for existing users.

## Summary by CodeRabbit

* **New Features**
  * Added privacy settings to control screenshot usage in dictation and app/window context sharing
  * Users can now skip screen recording permission during setup

* **UI/UX**
  * Settings view expanded with two new toggles explaining privacy tradeoffs
  * Setup flow now provides clear skip option for screenshot permissions, with disabled continue button until recording access is granted

[Review Change Stack](https://app.coderabbit.ai/change-stack/zachlatta/freeflow/pull/209?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/AppContextService.swift
+++ b/Sources/AppContextService.swift
@@ -1,6 +1,7 @@
 import Foundation
 import ApplicationServices
 import AppKit
+import os.log
 
 struct AppSelectionSnapshot {
     let appName: String?
@@ -36,6 +37,7 @@ Return only two sentences, no labels, no markdown, no extra commentary.
 """
     static let defaultContextPromptDate = "2026-02-24"
     static let defaultScreenshotMaxDimension: CGFloat = 1024
+    fileprivate static let contextLog = OSLog(subsystem: "com.zachlatta.freeflow", category: "Context")
 
     private let apiKey: String
     private let baseURL: String
@@ -409,7 +411,19 @@ Selected text: \(selectedText ?? "None")
         appElement: AXUIElement,
         focusedWindowTitle: String?
     ) -> (dataURL: String?, mimeType: String?, error: String?) {
+        let useScreenshotContext = UserDefaults.standard.object(forKey: "use_screenshot_context") == nil
+            ? true
+            : UserDefaults.standard.bool(forKey: "use_screenshot_context")
+        if !useScreenshotContext {
+            os_log(.info, log: Self.contextLog, "screenshot skipped — disabled in settings (Use screen context for smarter dictation = off)")
+            return (
+                nil,
+                nil,
+                "Screen context disabled in settings. Toggle it on in Settings → Prompts → Context Prompt to use screenshots."
+            )
+        }
         if !CGPreflightScreenCaptureAccess() {
+            os_log(.info, log: Self.contextLog, "screenshot blocked — Screen Recording permission not granted")
             return (
                 nil,
                 nil,
@@ -485,6 +499,7 @@ Selected text: \(selectedText ?? "None")
                     compression: screenshotCompressionPrimary,
                     maxDimension: screenshotMaxDimension
                 ) {
+                    os_log(.info, log: Self.contextLog, "screenshot captured via active-window match (%{public}d bytes)", dataURL.count)
                     return (dataURL, "image/jpeg", nil)
                 }
             }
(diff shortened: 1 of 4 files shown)


marcbodea:
same as #208, does not build due to missing ‘SettingsSubcard’
Sources/SettingsView.swift:1696:13: error: cannot find 'SettingsSubcard' in scope
1694 | 
1695 |         return VStack(alignment: .leading, spacing: 10) {
1696 |             SettingsSubcard {
     |             `- error: cannot find 'SettingsSubcard' in scope
1697 |                 VStack(alignment: .leading, spacing: 6) {
1698 |                     Toggle("Use screen context for smarter dictation", isOn: $appState.useScreenshotContext)
make: *** [build/FreeFlow Dev.app/Contents/MacOS/FreeFlow Dev] Error 1

PR #215, opened 2026-05-27
Show a subtle "still recording" sheen every 30s during long takes

ojhurst:
The mic waveform is small. When a recording is left running by accident — forgotten toggle, started by mistake — there isn't much in the menu bar to catch the eye. This adds a periodic sheen sweep across the recording overlay every 30 seconds: a quiet "yep, still here, still recording" that's easy to ignore when the recording is intentional and hard to miss when it isn't.

The sheen widens the overlay slightly, sweeps a soft highlight across left and right wings with a small stagger, then returns. No sound, no modal, no notification. The cadence resets when recording is toggled on, so the first sheen lands at the 30-second mark and repeats from there.

recording overlay sheen sweep every 30 seconds

### Implementation

- `AppState`: per-recording timer, reset on toggle start/stop
- `RecordingOverlay`: sheen sweep animation, staggered left/right wings, expansion and collapse around the recording overlay's shape

### Notes from the design pass

A few things considered along the way:

- **Cadence.** Shorter intervals (15 seconds) felt like a notification rather than an acknowledgement. 30 seconds reads as periodic reassurance, not a prompt.
- **Sheen-only.** Earlier iterations tried widening the overlay or pulsing it on its own. The sheen alone, with a subtle width expansion during the sweep, was the lightest touch that still caught the eye.
- **Brightness.** The peak opacity was tuned between "flashing alert" and "easy to miss." The committed value sits between — visible against the dark overlay without reading as urgent.
- **Pill variants and non-notched Macs.** The sheen geometry is anchored to the recording overlay's own bounds, not absolute pixel positions, so it scales naturally to wider pill layouts and to Macs without a notch — no per-layout code path needed.
- **First-sheen timing.** The 30-second timer resets on every recording toggle, so the first sheen always lands 30 seconds after start — never an immediate sheen on a fresh recording.
- **No sound, no modal.** The reminder is deliberately ignorable when intentional. If the recording is on purpose, the sheen never escalates; the goal is to draw the eye, not interrupt the work.

### Defaults

On by default. The sheen is gentle enough to leave running, and the 30-second cadence was tuned to feel like a periodic acknowledgement rather than a notification.

## Release Notes

* **New Features**
  * Enhanced recording session duration tracking and metrics
  * Toggle recording mode now includes visual reminders and notifications
  * Improved error and silence warning message display in the recording interface
  * Audio feedback on recording start
  * Visual enhancements to the recording overlay design

[Review Change Stack](https://app.coderabbit.ai/change-stack/zachlatta/freeflow/pull/215?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -527,6 +527,23 @@ final class AppState: ObservableObject, @unchecked Sendable {
         didSet {
             guard oldValue != isRecording else { return }
             AppState.writeRecordingStateFlag(isRecording)
+            // Build 119: capture per-session dictation duration so the WPM
+            // stat in MainWindow's right rail has real numerator/denominator
+            // material. Counts only the recording window; transcription and
+            // post-processing happen after isRecording flips back to false.
+            if isRecording {
+                recordingSessionStartTime = Date()
+            } else {
+                stopToggleRecordingReminder()
+                if let start = recordingSessionStartTime {
+                    let elapsed = Date().timeIntervalSince(start)
+                    if elapsed > 0 {
+                        wpmTrackedSeconds += elapsed
+                        WordCountTracker.saveWPMTrackedSeconds(wpmTrackedSeconds)
+                    }
+                    recordingSessionStartTime = nil
+                }
+            }
         }
     }
     @Published var isTranscribing = false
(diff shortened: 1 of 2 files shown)


ojhurst:
I have had this running in my daily driver now for a little while, and it's definitely helped me at least three different times. It's helped me catch the fact that I accidentally left it on.

marcbodea:
i'm not against showing a still recording reminder but not sure the sheen matches the rest of the app

ojhurst:
Have you tried the sheen? I feel like it's just a really subtle thing. Do you have any other mechanisms you would propose instead?

marcbodea:
Hi, could you make sure the sheen is restricted to the shape of the actual overlay? In the GIF you added it seems to be restricted to a square
image

Could you also address the coderabbit comments?

Thanks

PR #221, opened 2026-05-31
Mitigate security scan findings

mmmnorthmark:
## Summary

This PR hardens five areas surfaced by a security scan of FreeFlow. Each change is defensive and behavior-preserving for normal use; the only user-visible change is that local run-log retention is now opt-in (see below).

### 1. Provider URL validation & safe URL construction
- New `ProviderURLPolicy` centralizes base-URL validation: the URL must parse, use `http`/`https`, and include a host. URLs are normalized (trailing slashes trimmed).
- Request URLs are now built with `appendingPathComponent("chat")/("completions")` instead of `URL(string: "\(baseURL)/chat/completions")!`. This removes force-unwraps that could crash on a malformed base URL and avoids string-interpolation URL construction.
- Applied across `TranscriptionService`, `PostProcessingService`, `AppContextService`, and `RealtimeTranscriptionService`.

### 2. Store API keys in the Keychain instead of plaintext
- Secrets now live in the macOS Keychain with `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` rather than in a plaintext JSON settings file.
- Existing plaintext values are migrated into the Keychain and then removed from disk.

### 3. Stop leaking raw provider responses into errors/logs
- Provider error responses are no longer echoed verbatim into error messages or `os_log` output. Only the HTTP status code and response byte count are surfaced, avoiding accidental disclosure of sensitive response contents and log injection.

### 4. Local history retention is now opt-in
- Run-log / pipeline-history retention defaults to **off**. A "Keep Run Log" toggle in Settings enables it.
- When disabled, stored history entries and their associated audio files are cleared and not persisted, reducing local retention of potentially sensitive transcripts and audio.

### 5. Harden the auto-update path
Before installing a downloaded update, `UpdateManager` now verifies:
- the release contains the exact expected asset (`FreeFlow.dmg`);
- the download URL (initial **and** post-redirect) is on the expected GitHub release host;
- the downloaded size matches the asset's advertised size exactly;
- the app bundle identifier matches;
- the code signature is valid and the signing **team identifier** matches;
- the app passes a Gatekeeper assessment.

This prevents installing a tampered or spoofed update.

## Notes
- This branch is a single, focused commit on top of current `main`.
- No new dependencies.

## Summary by CodeRabbit

* **New Features**
  * Added a "Keep Run Log" setting to control pipeline history retention.

* **Bug Fixes**
  * Enhanced security validation for provider URLs and app updates.
  * Improved error messages for invalid provider configurations.
  * Strengthened update installation with code signing and Gatekeeper verification.

* **Improvements**
  * Better handling of provider URL normalization.
  * Refined error reporting for transcription and processing services.

### DIFF
--- a/Sources/AppContextService.swift
+++ b/Sources/AppContextService.swift
@@ -38,7 +38,7 @@ Return only two sentences, no labels, no markdown, no extra commentary.
     static let defaultScreenshotMaxDimension: CGFloat = 1024
 
     private let apiKey: String
-    private let baseURL: String
+    private let baseURL: URL?
     private let customContextPrompt: String
     private let contextModel: String
     private let maxScreenshotDataURILength = 500_000
@@ -57,7 +57,7 @@ Return only two sentences, no labels, no markdown, no extra commentary.
         screenshotMaxDimension: CGFloat = AppContextService.defaultScreenshotMaxDimension
     ) {
         self.apiKey = apiKey
-        self.baseURL = baseURL
+        self.baseURL = try? ProviderURLPolicy.normalizedHTTPBaseURL(from: baseURL)
         self.customContextPrompt = customContextPrompt
         let trimmedModel = contextModel.trimmingCharacters(in: .whitespacesAndNewlines)
         self.contextModel = trimmedModel.isEmpty ? "meta-llama/llama-4-scout-17b-16e-instruct" : trimmedModel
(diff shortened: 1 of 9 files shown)


zachlatta:
I don't think any of the security findings in this are legitimate so I'm closing. They all appear to be completely AI generated. If you want to open a handwritten PR with a demo video showing the changes you made, happy to consider.

PR #222, opened 2026-06-01
feat: easy drop down LLM model selection + add support for the qwen3-32b

pianoandglass:
This PR introduces a smoother and easier way to select and switch between LLM models in the Advanced Provider Settings, while adding robust support for the `qwen3-32b` reasoning model.

### What's New?

- **Easy Drop-Down Model Selection**: We replaced the simple text fields with a native macOS dropdown menu (Picker) for all model settings (Post-Processing, Fallback, Context, and Transcription). 
  - *Custom Models*: The dropdown includes a "Custom..." option that dynamically reveals a text field, ensuring power users can still manually type any model string they want without losing backwards compatibility.
- **`qwen3-32b` Support**: Added full support for Qwen reasoning models, which are exceptional at following complex custom system prompts. Previously, they couldn't be effectively used because their chain-of-thought `<think>` tags would leak into the final output. The post-processing pipeline now natively detects and cleanly strips these tags before outputting the final text.
- **Centralized Model Configuration**: Introduced `ModelConfiguration.swift` to act as the single source of truth for model capabilities. It allows us to manually tune parameters for specific models, such as `max_completion_tokens` or `reasoning_effort`. More importantly, this sets a strong foundation to make future model management and API payload adjustments much easier.

### Screenshots
SCR-20260601-orfm-2
SCR-20260601-otdn
SCR-20260601-osgu
SCR-20260601-orya

### Under the Hood

To implement this, we brought in `ModelConfiguration.swift` to act as our centralized hub for all model settings, including the new tag-stripping logic and reasoning effort configurations. Then, we created a brand new SwiftUI component called `ModelDropdownView.swift`, which serves as a highly reusable Picker that elegantly falls back to a text field for custom inputs. With those pieces in place, we updated the UI in `SettingsView.swift` to swap out the old text fields for our new dropdown component. Finally, we tied it all together by updating the `PostProcessingService.swift` to respect these new configurations, ensuring the API payloads are dynamically built and that any reasoning tags from models like Qwen are completely stripped out before reaching the final transcript.

## Summary by CodeRabbit

* **New Features**
  * Support for multiple LLM and transcription models with per-model settings and curated model lists
  * Dropdown model selector with a "Custom..." option and reset

* **Improvements**
  * Settings UI consolidated into streamlined model pickers for four model fields
  * Post-processing adapts behavior per selected model (completion tokens, reasoning options, optional stripping of <think> blocks)
  * Custom model entries commit on submit or when focus changes

### DIFF
--- a/Sources/ModelConfiguration.swift
+++ b/Sources/ModelConfiguration.swift
@@ -163,7 +163,7 @@ public struct ModelConfiguration {
         var cleaned = text
         
         // First, replace fully closed tags: <think>...</think>
-        let closedRegexPattern = "<think>[\\s\\S]*?</think>"
+        let closedRegexPattern = "^\\s*<think>[\\s\\S]*?</think>"
         if let regex = try? NSRegularExpression(pattern: closedRegexPattern, options: []) {
             let range = NSRange(cleaned.startIndex..<cleaned.endIndex, in: cleaned)
             cleaned = regex.stringByReplacingMatches(in: cleaned, options: [], range: range, withTemplate: "")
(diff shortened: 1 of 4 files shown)


marcbodea:
Great PR, thank you for this!

PR #225, opened 2026-06-06
In toggle mode, tap any one shortcut key (e.g. Fn or Space from Fn+Space) to end dictation — not the full combo

pianoandglass:
# Description

In **Tap to Toggle** mode, the dictation shortcut is a combination of two or more keys (e.g. `Fn` + `Space`, or `Fn` + `Control` + `Space`) that starts dictation with a single tap — no holding required. This PR changes how you **end** that dictation.

## Before → After

- **Before:** with a `Fn` + `Space` shortcut, ending dictation required pressing the **whole** `Fn` + `Space` combo again, or clicking the on-screen Stop button.
- **After:** with the same `Fn` + `Space` shortcut, tapping **either** `Fn` **or** `Space` on its own ends the dictation and sends it for transcription. This works for any multi-key combination (e.g. `Fn` + `Control` + `Space` → any one of the three keys ends it). Pressing `Space` to stop does **not** type a stray space into your text field.

## Changed Files

| File | Changes |
|---|---|
| `Sources/AppState.swift` | Implemented the `.consumeOnly` action to swallow trailing shortcut keys so they don't leak into the active text field. |
| `Sources/ShortcutCore/DictationShortcutSessionController.swift` | Added the `ignoreToggleUntilComponentDeactivation` flag to correctly track and consume stop-sequence events. Implemented the `.consumeOnly` action to prevent trailing keys from leaking. |
| `Sources/ShortcutCore/ShortcutMatcher.swift` | Added logic to detect individual toggle component inputs and emit stateless `toggleComponentInputReceived` events. Refactored modifier extraction to use safe `if let` binding instead of force unwrapping. |
| `Sources/ShortcutCore/ShortcutModels.swift` | Added the `toggleComponentInputReceived` event to the `ShortcutEvent` enumeration. |
| `Sources/HotkeyManager.swift` | Updated the `onShortcutEvent` handler to evaluate multiple emitted events per loop and correctly flag when a key press must be consumed by the OS. |
| `Sources/SetupTestHotkeyHarness.swift` | Updated the event closure signature to handle and consume the new `.consumeOnly` action. |
| `Sources/SetupView.swift` | Added support for the `.consumeOnly` action in the shortcut action switch statement. |

## Technical Details: Consuming Trailing Keys and Edge Cases

Handling global multi-key shortcuts requires carefully consuming keyboard events so they do not accidentally leak into the active text editor. The logic handles the following edge cases:

* **Consuming the Space Key (Trailing Keys)**: When a user stops dictation by releasing a sequence (like `Fn` + `Space`), they might release `Fn` before `Space`. This triggers a `toggleComponentInputReceived` event. The session controller intercepts this stateless event and returns `.consumeOnly`, which tells macOS to completely swallow the keystroke, preventing a stray space from being typed into the text box.
* **Component-Level Deactivation**: To prevent the keyboard from getting permanently locked, the system tracks when all components of the shortcut are physically released (`toggleComponentDeactivated`). It only resets the ignore state when the user completely lifts their fingers off the keys, preventing accidental double-triggers.

## Release Notes

* **New Features**
  * Improved hotkey behavior with clearer “consumed vs. passed through” handling.
  * Added finer-grained toggle shortcut lifecycle tracking, including component-level activation, input, and deactivation.

* **Bug Fixes**
  * Refined event consumption to prevent unintended keyboard input propagation.
  * Improved dictation toggle responsiveness, including safer handling during rapid start/stop and state transitions.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -1612,12 +1612,11 @@ final class AppState: ObservableObject, @unchecked Sendable {
         return false
     }
 
+    // Starts monitoring system-wide hotkeys and binds event handlers
     func startHotkeyMonitoring() {
         shouldMonitorHotkeys = true
         hotkeyManager.onShortcutEvent = { [weak self] event in
-            DispatchQueue.main.async {
-                self?.handleShortcutEvent(event)
-            }
+            self?.handleShortcutEvent(event) ?? false
         }
         hotkeyManager.onEscapeKeyPressed = { [weak self] in
             self?.handleEscapeKeyPress() ?? false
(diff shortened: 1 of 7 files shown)


pianoandglass:
Rebased onto the latest `main` — this PR is now **conflict-free and cleanly mergeable** against current upstream.

It's intentionally **self-contained**: scoped to compile standalone against `main` with a minimal footprint on shared files, so it does not depend on — or conflict with — my other open PRs. Each can be reviewed and merged independently, in any order.

@coderabbitai review

PR #226, opened 2026-06-06
Add option to paste custom words into vocabulary from status bar

pianoandglass:
# Add option to paste custom words into vocabulary from status bar

This change adds a button to the status bar menu to quickly append clipboard content directly to the custom vocabulary. This allows users to add new terms without opening the settings window.

## Proposed Changes

- Added a "Paste Custom Word to Vocabulary" option in the status bar menu.
- Implemented vocabulary updating logic via an extension on AppState to retrieve, sanitize, and append clipboard text to the custom vocabulary, preserving any intentional spacing.
- Created a transient notification manager that displays a visual checkmark indicator next to the status bar icon for 2 seconds when a word is successfully added.
- The feature ensures Swift 6 concurrency safety by fully adopting `@MainActor` and `@unchecked Sendable` where applicable.

## Modified and Added Files

| File | Status | Description |
| :--- | :--- | :--- |
| [Sources/AppState+AddVocabularyButton.swift](file:///Users/saphi/bin/freeflow%202/Sources/AppState%2BAddVocabularyButton.swift) | Added | Extension containing the logic to fetch text from the pasteboard and append it to the custom vocabulary, preserving internal spacing and avoiding duplicates. |
| [Sources/VocabularyNotificationManager.swift](file:///Users/saphi/bin/freeflow%202/Sources/VocabularyNotificationManager.swift) | Added | Manager to handle the transient state of the status bar notification securely isolated to the Main Actor. |
| [Sources/App.swift](file:///Users/saphi/bin/freeflow%202/Sources/App.swift) | Modified | Updated the status bar label to display a checkmark when a vocabulary word is successfully added, adopting Swift 6 concurrency rules. |
| [Sources/MenuBarView.swift](file:///Users/saphi/bin/freeflow%202/Sources/MenuBarView.swift) | Modified | Added the "Paste Custom Word to Vocabulary" button to the status bar menu. |

## Summary by CodeRabbit

* **New Features**
  * Added "Paste Custom Word to Vocabulary" menu button for quick clipboard import
  * Clipboard content is validated, trimmed, de-duplicated, and supports multiple words/lines
  * Menu bar now shows an animated checkmark confirmation for 2 seconds on successful paste

### DIFF
--- a/Sources/App.swift
+++ b/Sources/App.swift
@@ -17,8 +17,10 @@ struct FreeFlowApp: App {
     }
 }
 
+@MainActor
 struct MenuBarLabel: View {
     @EnvironmentObject var appState: AppState
+    @ObservedObject var notificationManager = VocabularyNotificationManager.shared
 
     private var iconName: String {
         if appState.isRecording { return "record.circle" }
@@ -27,12 +29,18 @@ struct MenuBarLabel: View {
     }
 
     var body: some View {
-        if AppBuild.isDevBundle && !appState.isRecording && !appState.isTranscribing {
-            Image(nsImage: StampedMenuBarIcon.templateImage)
-                .renderingMode(.template)
-        } else {
-            Image(systemName: iconName)
+        HStack(spacing: 4) {
+            if notificationManager.showCheckmark {
+                Image(systemName: "checkmark")
+            }
+            if AppBuild.isDevBundle && !appState.isRecording && !appState.isTranscribing {
+                Image(nsImage: StampedMenuBarIcon.templateImage)
+                    .renderingMode(.template)
+            } else {
+                Image(systemName: iconName)
+            }
         }
+        .animation(.easeInOut(duration: 0.2), value: notificationManager.showCheckmark)
     }
 }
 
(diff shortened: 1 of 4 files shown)


pianoandglass:
I adjusted the tokenization of `pastedString` and the deduplication logic to correctly split them by commas, semicolons, and line breaks. I also added a `.trimmingCharacters` step to each token before filtering and comparison, ensuring full compatibility with `PostProcessingService.mergedVocabularyTerms` and fixing the false negative issue in deduplication.

marcbodea:
Thanks for this!

I love wispr flow's implementation of "auto-detecting" changes, not sure how we could implement it but would be interesting to take a look. In the meantime will test and merge this

pianoandglass:
> Thanks for this!
> 
> I love wispr flow's implementation of "auto-detecting" changes, not sure how we could implement it but would be interesting to take a look. In the meantime will test and merge this

I love it too! In fact, I already worked on a first attempt about a month ago, even implementing a fancy animation that shows the term/word added in the OSD, and it was working well most of the time. However, several intentional text changes were causing false positives, resulting in many things being added randomly. I want to try a new approach for implementing that.

In any case, one thing doesn't exclude the other, because there can be terms that the user wants to add intentionally that aren't necessarily transcribed incorrectly.

This PR works great and is very intuitive! Thanks for the feedback @marcbodea  :)

PR #228, opened 2026-06-09
fix: guard against executed dictation prompts

### LINKED ISSUE #187: Voice instructions sometimes get executed instead of being transcribed
Hello,

I encountered this problem while asking to reply to an email in a different language than English. And more often than not, I got an execution of the instructions rather than a transcription.

_Example voice prompt:_
```
Please write me an email in Lithuanian that would respond for my client saying I'm very sorry that this issue happened to you, we are working relentlessly to solve the issue and we will get back to you as soon as possible.
```

_The result I sometimes get:_
```
Labai atsiprašome, kad šis klausimas kilo jums. Mes labai intensyviai dirbame, kad jį išspręstume, ir greičiausiai susisieksime su jumis.
```

Also, I wish there **was a button to opt out from post processing** and just get the raw transcription. I did this manually by just entering dashes for the post-processing model name paths.

All in all, it's a terrific app and I'm very grateful to have it!

Best,
Mindaugas

Tektongrowth:
## Summary
- Treat the raw transcript in post-processing as delimited data instead of inline prompt text
- Detect likely cases where the cleanup model answered/executed the dictated text instead of cleaning it
- Fall back to the retry model, then to the literal trimmed transcript, rather than pasting an AI-generated answer

## Why
Normal dictation can contain text that sounds like a prompt, for example:

> Please write me an email in Lithuanian that would respond for my client saying I am sorry this happened.

In those cases, the cleanup model can sometimes follow the dictated instruction and return a generated answer. FreeFlow then pastes that answer instead of the literal transcript.

This PR hardens the post-processing path so the safer failure mode is an unpolished transcript, not an executed instruction.

## Verification
- Built successfully on macOS ARM64 with:

```bash
make APP_NAME="FreeFlow Dev" CODESIGN_IDENTITY="-"
```

- Manually tested the reported failure class with `FreeFlow Dev.app`; the dictated instruction-like text was pasted as literal transcript rather than executed/generated output.

Fixes #187

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transcript post-processing to detect when output appears to execute dictated instructions rather than being cleaned, reducing unnecessary retries.
  * Updated fallback behavior to return a safer result when this condition is detected.
* **New Features**
  * Added an “Instruction Guard” toggle in System Prompt settings to control whether the app retries/falls back to the literal transcript in suspicious cases.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -217,6 +217,7 @@ final class AppState: ObservableObject, @unchecked Sendable {
     private let selectedMicrophoneStorageKey = "selected_microphone_id"
     private let customSystemPromptStorageKey = "custom_system_prompt"
     private let customContextPromptStorageKey = "custom_context_prompt"
+    private let instructionExecutionGuardEnabledStorageKey = "instruction_execution_guard_enabled"
     private let customSystemPromptLastModifiedStorageKey = "custom_system_prompt_last_modified"
     private let customContextPromptLastModifiedStorageKey = "custom_context_prompt_last_modified"
     private let contextScreenshotMaxDimensionStorageKey = "context_screenshot_max_dimension"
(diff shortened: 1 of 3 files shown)


marcbodea:
Added a settings toggle in case this breaks some specific use cases

PR #230, opened 2026-06-11
Stop permission polling timer once both permissions are granted (to save CPU and battery)

ojhurst:
## The story

A battery monitor flagged FreeFlow Dev sitting at ~10% CPU on a MacBook Pro — with the app idle, no recording, no dictation active. Profiling showed two system calls firing every 2 seconds for the entire lifetime of the process.

The trail led to `startAccessibilityPolling()` in `AppState.swift`. It installs a repeating `Timer` that checks `AXIsProcessTrusted()` and `CGPreflightScreenCaptureAccess()` every 2 seconds — both IPC calls to `tccd` (the TCC privacy daemon). The timer is started from `applicationDidFinishLaunching()` and `completeSetup()`, but there is no code path that stops it during normal operation. `stopAccessibilityPolling()` is only called when re-entering setup.

The polling exists so the setup wizard can detect when the user grants a permission in System Settings. That is a reasonable use during setup. After setup, both permissions are granted and stable — they only change if the user manually revokes them in System Settings, which is extremely rare.

## The bug

`startAccessibilityPolling()` unconditionally installs a 2-second repeating timer. The timer fires 30 times per minute, making 60 IPC calls per minute to `tccd` and writing to two `@Published` properties on each tick. Those writes trigger SwiftUI's observation system and invalidate `MenuBarView` (which reads both `hasAccessibility` and `hasScreenRecordingPermission`) on every tick — even when the values have not changed.

## The fix

Two changes to `startAccessibilityPolling()`:

1. **Early return when already granted.** After the initial permission check, if both are `true`, return without installing the timer. This is the common case after first launch.

2. **Auto-stop on grant.** If the timer does start (because a permission was missing during setup), each tick checks whether both permissions are now granted. When they are, the timer invalidates itself and nils out.

## Behavior

- **Normal launch (permissions already granted):** Timer never starts. Zero ticks, zero IPC, zero SwiftUI invalidation.
- **First launch / setup (permissions not yet granted):** Timer runs as before until both are granted, then auto-stops.
- **Re-run setup:** `stopAccessibilityPolling()` / `startAccessibilityPolling()` cycle is unchanged.
- **Permission revoked mid-session:** The cached `@Published` value may go stale. The `prepareRecordingStart()` guard catches this at recording time and shows the accessibility alert. In practice this is identical to current behavior — the 2-second poll provided at most a 2-second head start in detecting a revocation that the user just performed themselves.

## Testing performed

Tested on MacBook Pro (M3 Pro, 16 GB) running macOS 15.5.

| Scenario | Result |
|----------|--------|
| Launch with both permissions granted | Timer skipped, log confirms `both permissions granted at launch — polling skipped` |
| Normal dictation after fix | Works as before — hold hotkey, dictate, release, text inserted |
| Menu bar dropdown | All items render correctly, accessibility/screen recording warnings hidden (as expected when granted) |
| Multiple relaunches over 2 hours | Every launch shows polling skipped, zero timer ticks across all launches |

**Instrumentation:** Added `os_log` at default level under category `Permissions` to confirm the decision path. Seven separate launches over 2 hours all showed `polling skipped`. No instance of the timer starting.

## Test plan for reviewer

- [ ] Build and launch on a Mac with both permissions already granted — verify dictation works normally
- [ ] `log show --predicate 'process == "FreeFlow Dev" AND category == "Permissions"' --last 5m` should show `both permissions granted at launch — polling skipped`
- [ ] Reset TCC: `tccutil reset Accessibility com.zachlatta.freeflow.dev` — relaunch, verify the setup wizard still detects the grant
- [ ] Re-run setup from Settings, verify the polling cycle restarts and stops correctly

## Notes / Considered

- **Why not use KVO or `DistributedNotificationCenter` to detect permission changes instead of polling?** macOS does not provide a public notification for TCC permission changes. `AXIsProcessTrusted()` and `CGPreflightScreenCaptureAccess()` are the documented APIs, and they are poll-only. The alternative would be `CGSSessionNotificationRegister` (SPI), which is fragile and undocumented.
- **Why not reduce the interval instead of stopping?** Even a 30-second poll is unnecessary when the state changes once per machine lifetime. The correct interval for "never changes" is "do not poll."
- **Could `@Published` property writes be deduplicated instead?** SwiftUI's `@Published` does not deduplicate — every write triggers `objectWillChange`. A `willSet` guard could skip writes when the value has not changed, but that treats a symptom. The timer itself is the waste.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accessibility and screen-recording permission handling to avoid unnecessary background polling when permissions are already granted, reducing resource use.
  * Permission status is refreshed immediately when preparing to record and polling stops promptly once permissions become available, making recording start more reliable.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -1241,12 +1241,21 @@ final class AppState: ObservableObject, @unchecked Sendable {
 
     func startAccessibilityPolling() {
         accessibilityTimer?.invalidate()
+        accessibilityTimer = nil
         hasAccessibility = AXIsProcessTrusted()
         hasScreenRecordingPermission = hasScreenCapturePermission()
+        if hasAccessibility && hasScreenRecordingPermission {
+            return
+        }
         accessibilityTimer = Timer.scheduledTimer(withTimeInterval: 2.0, repeats: true) { [weak self] _ in
             DispatchQueue.main.async {
-                self?.hasAccessibility = AXIsProcessTrusted()
-                self?.hasScreenRecordingPermission = self?.hasScreenCapturePermission() ?? false
+                guard let self else { return }
+                self.hasAccessibility = AXIsProcessTrusted()
+                self.hasScreenRecordingPermission = self.hasScreenCapturePermission()
+                if self.hasAccessibility && self.hasScreenRecordingPermission {
+                    self.accessibilityTimer?.invalidate()
+                    self.accessibilityTimer = nil
+                }
             }
         }
     }
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -1925,7 +1925,9 @@ final class AppState: ObservableObject, @unchecked Sendable {
         startedAt: CFAbsoluteTime? = nil
     ) -> Bool {
         activeRecordingTriggerMode = triggerMode
-        guard hasAccessibility else {
+        let isAccessibilityTrusted = AXIsProcessTrusted()
+        hasAccessibility = isAccessibilityTrusted
+        guard isAccessibilityTrusted else {
             errorMessage = "Accessibility permission required. Grant access in System Settings > Privacy & Security > Accessibility."
             statusText = "No Accessibility"
             activeRecordingTriggerMode = nil


marcbodea:
great PR, thank you!

PR #231, opened 2026-06-12
User-configurable 'Press Enter' voice command — custom & multilingual trigger phrases

pianoandglass:
## Description

FreeFlow already had a single, hard-coded **"press enter"** trigger that presses Return after a dictation. This PR makes the trigger **fully user-configurable**.

### What it adds
- A **Settings UI** to add and remove your own trigger phrases as removable tags — in **any language** and in **custom wordings** (e.g. "hit enter", "pressione enter", "send", "appuyer sur Entrée" — whatever you want).
- A quick-add **library of multilingual variations** with built-in suggestions in the UI.
- A robust extraction engine that strips the matched trigger from the **end** of the transcript — even across **repeated** triggers, **multi-line** endings, and trailing punctuation — **before** the text reaches the LLM cleanup step, **recording the intent to press Enter right after the cleaned text is pasted**.

---

SCR-20260616-ovpj
---
SCR-20260616-ouqr

---

### UI and UX

- **Tag-Based Settings Interface**: Overhauled the Press Enter settings section. Users can now view active variations as persistent UI tags.
- **Suggested Variations**: Integrated a quick-add library containing built-in variations for some common languages.
- **Apple HIG Compliance**: Active and suggested tags are designed with native macOS characteristics, including capsule shapes, semi-transparent backgrounds (`Color.accentColor.opacity(0.15)`), and smooth hover states.
- **SwiftUI Performance Optimization**: Hover state logic (`@State`) for tags was extracted into independent structures (`PressEnterActiveTagView` and `PressEnterSuggestedTagView`). This isolation prevents the primary `SettingsView` from continuously re-rendering during pointer interactions, ensuring high application performance.

### Technical Implementation

- **Modular Regex Engine**: Created `MultilanguagePressEnter.swift` to handle command compilation and extraction. The system compiles user-defined active commands into a single, optimized, case-insensitive `NSRegularExpression`.
- **Advanced Suffix Extraction**: Replaced simple `.hasSuffix()` checks. The new engine correctly strips trigger phrases from the end of the transcript even when followed by punctuation (e.g., "?", "!"). Relevant punctuation is preserved and re-applied to the finalized transcript.
- **Consecutive Trigger Handling**: The extraction loop dynamically handles repeated trigger words (e.g., "final phrase. enter, enter"), stripping them consecutively until only the intended transcript remains.
- **Input Sanitization**: Automatically normalizes user input for custom variations by collapsing excess whitespaces to ensure robust regex compilation.

## Summary by CodeRabbit

* **New Features**
  * Added a customizable “Recognize ‘Press Enter’” setting in General Settings.
  * Users can now add, view, and remove their own press-enter phrases, with suggested phrases shown for quick setup.
  * The app now better recognizes multiple phrase variations and handles trailing punctuation more naturally.

* **Bug Fixes**
  * Improved command detection so spoken “press enter” variations are stripped more reliably from transcripts.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -225,6 +225,8 @@ final class AppState: ObservableObject, @unchecked Sendable {
     private let preserveClipboardStorageKey = "preserve_clipboard"
     private let keepDictationInClipboardHistoryStorageKey = "keep_dictation_in_clipboard_history"
     private let pressEnterVoiceCommandStorageKey = "press_enter_voice_command_enabled"
+    /// UserDefaults key under which the user's custom "Press Enter" command variations are persisted.
+    private let customPressEnterCommandsStorageKey = "custom_press_enter_commands"
     private let alertSoundsEnabledStorageKey = "alert_sounds_enabled"
     private let soundVolumeStorageKey = "sound_volume"
     private let voiceMacrosStorageKey = "voice_macros"
@@ -277,9 +279,6 @@ final class AppState: ObservableObject, @unchecked Sendable {
     static let defaultPostProcessingModel = "openai/gpt-oss-20b"
     static let defaultPostProcessingFallbackModel = "meta-llama/llama-4-scout-17b-16e-instruct"
     static let defaultContextModel = "meta-llama/llama-4-scout-17b-16e-instruct"
-    private static let trailingPressEnterCommandPattern = try! NSRegularExpression(
-        pattern: #"(?i)(?:^|[ \t\r\n,;:\-]+)press[ \t\r\n]+enter[\s\p{P}]*$"#
-    )
 
     @Published var hasCompletedSetup: Bool {
         didSet {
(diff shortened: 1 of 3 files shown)


pianoandglass:
Thanks @coderabbitai! All 3 issues have been addressed: 
- Fixed duplicate punctuation restoration logic
- Added \b word boundaries to regex matching to prevent sub-word matching
- Refactored ActiveTag remove button for keyboard accessibility and fixed syntax errors

pianoandglass:
Pushed a small follow-up commit (`059f36a`) that makes the new `compiledRegex` parameter of `parseTranscriptCommands(...)` optional (`NSRegularExpression? = nil`), so existing two-argument callers stay source-compatible.

`MultilanguagePressEnter.extractCommand(from:withRegex:)` already treats a nil regex as a no-op failsafe (returns the transcript unchanged), so there is no behavior change — normal dictation still passes the compiled regex explicitly. This keeps the function backward-compatible for callers that do not precompile a regex.

@coderabbitai review

pianoandglass:
Rebased onto the latest `main` — this PR is now **conflict-free and cleanly mergeable** against current upstream.

It's intentionally **self-contained**: scoped to compile standalone against `main` with a minimal footprint on shared files, so it does not depend on — or conflict with — my other open PRs. Each can be reviewed and merged independently, in any order.

@coderabbitai review

PR #246, opened 2026-06-28
fix: fall back to the backup cleanup model the moment the main one is rate-limited, instead of retrying it every dictation

pianoandglass:
Supersedes #238. That PR was reworked from scratch after a self-review surfaced edge cases in the 429 handling and to shrink the footprint in `PostProcessingService.swift`. Opening a fresh PR so the review starts clean on the final code.

## Summary

When the post-processing (cleanup) model returns HTTP 429 (rate limit), the app now records a short per-model cooldown and routes the next requests to the backup model up-front, instead of retrying the rate-limited model on every dictation. Daily limits are persisted and shown in Settings so you can see when a model is unavailable until its quota resets.

## Changes

| Component | Details |
|---|---|
| `LLMCooldownManager.swift` (new) | Small actor that records per-model cooldowns — minute-level in memory, daily (>= 1h) persisted to UserDefaults. Also owns the 429 header parsing (`retryAfterSeconds` / `parseGroqDuration`) and the up-front model choice (`effectivePrimary`). Self-contained so the backbone stays thin. |
| `PostProcessingService.swift` | Minimal hooks only: on a 429, register the cooldown and surface a `.rateLimited` error; before each request, switch to the backup if the primary is cooling down. |
| `SettingsView.swift` | A per-model daily-limit warning that appears when a limit is hit and auto-clears as the window resets. |

## What was wrong before (and is now fixed)

- Daily limits were never detected: only the per-minute reset header was read, and the duration parser only understood a bare `"<n>s"`, so daily limits never crossed the persistence threshold or showed in Settings. The daily-reset header is now read and compound formats (`2m59.56s`, `1h0m0s`, `120ms`) are parsed.
- The backup model's own rate limit was never recorded; now both the primary and the backup feed the breaker.
- The Settings warning did not refresh on its own; it now appears and auto-clears within a few seconds.

## Edge cases considered

- Malformed, negative, or non-finite header values are rejected, so a bad header can never produce a negative, NaN, or infinite cooldown.
- No backup configured, or both models rate-limited: the request falls through gracefully without retrying the same model.
- Header lookup is case-insensitive; when no timing header is present, a short re-probe cooldown is used.

## Testing

- The duration parser was unit-tested across bare, suffixed, compound, and invalid inputs (including negative / NaN / infinite, all rejected).
- Built and run in the integrated local build alongside the other in-progress features; no conflicts.

## Summary by CodeRabbit

* **New Features**
  * Added automatic cooldown tracking for rate-limited models to reduce repeated 429 retries.
  * The app now routes requests to a fallback model while the primary is cooling down.
  * Settings displays “daily limit reached” warnings with clearer reset times and automatically hides them after expiry.
* **Bug Fixes**
  * Improved retry and fallback behavior by using rate-limit headers to determine cooldown duration and ensuring safe fallback selection.
  * Enhanced rate-limit messaging to report the affected model and retry-after time.

### DIFF
--- a/Sources/LLMCooldownManager.swift
+++ b/Sources/LLMCooldownManager.swift
@@ -60,14 +60,17 @@ actor LLMCooldownManager {
         }
     }
 
-    /// Returns the model to use up-front: the fallback when the primary is in a rate-limit
-    /// cooldown and a fallback exists, otherwise the primary. Lets a caller route around a
-    /// cooling-down model in a single call. Behaviorally identical to the inline swap it replaces.
-    func effectivePrimary(_ primary: String, fallback: String?) -> String {
-        if isInCooldown(primary), let fallback {
-            return fallback
+    /// Returns an available model to use up-front, or nil when none is: the primary if it is not
+    /// cooling down; otherwise the fallback if it exists and is itself not cooling down; otherwise
+    /// nil, so the caller can skip a doomed request when BOTH models are rate-limited.
+    func effectivePrimary(_ primary: String, fallback: String?) -> String? {
+        if !isInCooldown(primary) {
+            return primary
         }
-        return primary
+        guard let fallback, !isInCooldown(fallback) else {
+            return nil
+        }
+        return fallback
     }
 
     // MARK: - UserDefaults (daily-limit persistence)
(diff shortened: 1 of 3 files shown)


pianoandglass:
@coderabbitai Thanks for the review. Pushed a commit addressing it:

**Fixed**

- **Daily-limit persistence** (`LLMCooldownManager.swift`): `rateLimitCooldown(from:)` now also reports whether the value came from the daily `x-ratelimit-reset-requests` (RPD) header, and `setCooldown(..., persist:)` persists those even when the reset is under an hour. A near-reset daily quota now survives restart and shows in Settings instead of staying in memory.
- **Raw-transcript safe-exit** (`PostProcessingService.swift`): split the retry guard so a `.suspectedInstructionExecution` on the first attempt after an up-front cooldown swap still returns the safe raw transcript instead of throwing, matching the explicit-retry path.
- **Reset-date formatting** (`SettingsView.swift`): added `formattedCooldownReset(_:)` that includes the date when the reset falls on a later day, so a cross-midnight daily reset is no longer an ambiguous bare time.

**Not changed (with reason)**

- **"Don't select a fallback already in cooldown" (`effectivePrimary`)**: left as-is by design. When *both* models are cooling down, the current code degrades gracefully — it sends one request, records that cooldown, and fails — which is the same outcome as before this breaker existed (and the common case, a cooling primary with a healthy fallback, is already handled). The suggested fail-fast requires returning an optional plus a new "no model available" path threaded through `PostProcessingService`, which is the protected backbone we are deliberately keeping minimal in this PR. Scoped out intentionally rather than expand the core file.
- **Docstring coverage warning**: every new declaration introduced by this PR has a `///` doc comment. The residual percentage is from pre-existing declarations in the touched backbone file (`PostProcessingService.swift`), which this PR intentionally does not refactor to avoid widening the diff into protected core code.

pianoandglass:
@coderabbitai Both points addressed in the latest commit:

- **Detect exhausted RPD before honoring `retry-after`** (`rateLimitCooldown`): it now checks `x-ratelimit-remaining-requests <= 0` first and, when the daily request quota is spent, returns the `x-ratelimit-reset-requests` reset with `isDaily = true` — before honoring `retry-after`. A short, near-reset daily window is now persisted and shown in Settings instead of being kept only in memory.
- **Don't return a fallback that is already cooling down** (`effectivePrimary`): it now returns `String?` — the primary if it isn't cooling, otherwise the fallback only if it isn't cooling either, otherwise `nil`. The two call sites skip the doomed request on `nil` and degrade gracefully: transcript cleanup returns the raw transcript, and Edit Mode returns the selection unchanged.

Thanks!

pianoandglass:
@coderabbitai review

PR #247, opened 2026-06-29
Chameleon — context-aware dictation that fits where you drop it (Beta - WIP)

### LINKED ISSUE #200: Spacing, capitalization, and punctuation are context-blind — no awareness of text surrounding the cursor
## Problem

FreeFlow currently has no awareness of the text surrounding the cursor when pasting dictated text. The spacing logic in `writeTranscriptToPasteboard` blindly appends a space after sentence-ending punctuation (`.!?`), and the LLM is not given any surrounding text context. This causes several issues:

### 1. Trailing whitespace artifact
Every dictation that ends with punctuation appends a trailing space, even when the cursor is at the end of a text field (no text follows). The user sees a visible trailing space after every sentence they dictate.

**Current behavior:** `"Hello world." ` (with trailing space)
**Expected behavior:** `"Hello world."` (no trailing space; consecutive dictations should add leading space instead)

### 2. Incorrect capitalization when inserting mid-sentence
The LLM always capitalizes the first word of the transcript because it has no knowledge of what comes before the cursor. When inserting text mid-sentence, this produces incorrect output.

**Example:** Cursor is after `"I think "` → dictates "we should go" → gets `"We should go"` (uppercase W)
**Expected:** `"we should go"` (lowercase, since we're mid-sentence)

### 3. Capitalization fails for list items
After a Markdown/plaintext list marker like `"- "`, `"* "`, or `"1. "`, the first letter should be capitalized, but the system doesn't detect list boundaries.

### 4. Duplicate/conflicting punctuation when inserting mid-text
The LLM adds terminal punctuation assuming the transcript ends the sentence. When the text following the cursor already has punctuation, this creates duplicates like `".."`  or conflicts like `".?"`.

### 5. No surrounding context sent to LLM
The post-processing LLM receives only `CONTEXT` (app activity summary) and `RAW_TRANSCRIPTION`, but never the text before and after the cursor. This means the LLM cannot make informed grammatical or spelling decisions based on the actual insertion point.

## Proposed Solution

1. **Read surrounding text via Accessibility API** — Use `kAXValueAttribute` + `kAXSelectedTextRangeAttribute` to read `precedingText` and `followingText` from the focused text field.
2. **Apply deterministic formatting rules** — Capitalize at sentence boundaries, newlines, and list markers; lowercase mid-sentence; remove conflicting punctuation; clean all Unicode whitespace variants.
3. **Smart spacing** — Leading space only when `precedingText` ends with a word character; trailing space only when `followingText` starts with a word character; no trailing space at end of field.
4. **Feed context to LLM** — Include `PRECEDING_TEXT` and `FOLLOWING_TEXT` in the user prompt so the model can make better cleanup decisions.
5. **Electron/Web hack** — For Chromium-based apps where the AX tree doesn't auto-sync, perform a zero-net-movement keypress (→←) to force refresh.
6. **Skip screenshot when text context is available** — Saves ~200ms per dictation when the Accessibility API can read the text field directly.

pianoandglass:
# Chameleon — context-aware dictation that fits where you drop it

*Dictated text now adapts to its surroundings — capitalization, spacing, and punctuation — instead of being pasted in raw at the cursor. It reads the text around the cursor without ever disturbing it, and the cleanup model uses that context as a hint too. The new post-LLM step happens entirely locally and instantly.*

Closes #200.

---

## What it does, in plain terms

Today, when you dictate, the cleaned-up text is dropped at the cursor exactly as the model produced it: the first word capitalized, a space tacked onto the end, and nothing else adjusted for what is actually around the cursor. That is fine when you are starting a clean new line — but the moment you are anywhere else (mid-sentence, right before a question mark, inside parentheses, in the middle of a list, or replacing a selection) the result comes out messy: a stray capital letter, missing or doubled spaces, and punctuation that clashes with what was already there.

This change reads the few characters around the cursor and reshapes the inserted text so it reads naturally in that exact spot: it fixes the first-letter case, adds or removes the space on each side, cleans up clashing punctuation, recognizes real paragraph breaks, and keeps your custom spellings. 

It works in normal Mac apps and inside web / Electron apps (chat apps, editors, browsers), where the cursor information is normally unreliable.

---

## See it in action

In each pair, `│` is the cursor. "Before" is how it works today (first word capitalized, a trailing space, nothing else touched); "After" is with this change.

**Mid-sentence** — at `I think we should │`, you say *"go home now"*
- Before: `I think we should Go home now.` *(stray capital G, plus a leftover space at the end)*
- After: `I think we should go home now.`

**Right before a question mark** — at `Did you finish│?`, you say *"the report yesterday"*
- Before: `Did you finishThe report yesterday. ?` *(glued to the word, capital T, a period, a space, then the ?)*
- After: `Did you finish the report yesterday?`

**In the middle of a list** — at `I bought apples│, oranges, and pears.`, you say *"bananas"*
- Before: `I bought applesBananas. , oranges, and pears.` *(glued, capital B, a period and a space dumped before the comma)*
- After: `I bought apples bananas, oranges, and pears.`

**Right before an existing comma** — at `we met on Tuesday│, then left`, you say *"early"* and the model adds its own comma/period
- Before: `we met on Tuesday Early., then left` *(the model's trailing mark collides with the comma already there)*
- After: `we met on Tuesday early, then left` *(the document's comma wins; the model's trailing mark is dropped)*

**Two sentences in a row** — at `I finished the task.│`, you say *"let's review it tomorrow"*
- Before: `I finished the task.Let's review it tomorrow.` *(no space after the first period, leftover space at the end)*
- After: `I finished the task. Let's review it tomorrow.`

**Inside a line of code** — at `print(│)`, you say *"hello world"*
- Before: `print(Hello world. )` *(capital H, a period, and a space before the closing parenthesis)*
- After: `print(hello world)`

**Replacing a selection** — in `the old replace`, you select `old` and say *"seamless"*
- Before: `the Seamless.  replace` *(capital S, a stray period, and a double space)*
- After: `the seamless replace`

**Custom vocabulary awareness, mid-sentence** — vocabulary `GitHub, API`; at `we use │ daily`, you say *"github and the api"*
- Before: `we use Github and the api.  daily` *(wrong casing, a period mid-sentence, a double space)*
- After: `we use GitHub and the API daily`

---

## What the Run Log shows

This feature also turns FreeFlow's Run Log into a full, readable trace of every dictation, with a completely new stage 4 of the pipeline, logging exactly how a piece of text travelled from your voice to the cursor. Most of this is new:

- **The new fourth stage: deterministic, on-device formatting.** After the model, the text runs through the rule-based formatter this change adds. The Run Log shows its formatted output and a plain-English summary of every rule that fired — no network, no model, just deterministic local rules.
- **What was read around the cursor.** The kind of app (a standard Mac app or a web/Electron view), how the surrounding text was read (the exact method/rung that succeeded), whether any text was found at all, and the actual text before and after the cursor — the exact context the insertion was based on.
- **What the post-processing model received and returned.** The raw transcript that went into the cleanup model and the cleaned text it gave back, shown together with the bounded surrounding context that informed it.
- **How it was finally inserted.** The resulting text shown in place (text before + inserted text + text after), exactly as it landed at the cursor.

The labels are faithful: the "how it was read" line reflects the method that actually produced the context (so if the non-destructive start-of-recording snapshot was used, that is what it says — not a stale value).

That summary of the fourth stage reads like this:

- "Capitalized the first letter" — at the start of an empty field.
- "Made the first letter lowercase, added a space before" — continuing a sentence mid-line.
- "Capitalized the first letter, added a space before" — right after a period.
- "Added spaces on both sides" — dropped between two words.
- "Added a space after" — inserted just before existing text.
- "Removed an extra punctuation mark" — a period/comma/semicolon clashing with a following `)`, `?`, or comma.
- "Removed an extra punctuation mark, made the first letter lowercase" — a mid-sentence period before lowercase text.
- "Restored a custom-word spelling" — a vocabulary word corrected (github to GitHub).
- "Replaced the selected text" — dictating over a selection.
- "No changes" — the text already fit perfectly.

---

## What's new

1. **Context-aware capitalization.** The first letter follows the sentence: uppercase at the start of a field, after `.`/`?`/`!`/`…`, after a line/paragraph break, after a bullet, or when opening a quote/parenthesis that begins a sentence — lowercase when you are continuing one. Abbreviations (`Dr.`, `i.e.`, `etc.`-shaped tokens), the pronoun `I` and its contractions (`I'm`, `I'll`…), and all-caps acronyms (`API`, `NASA`) are never mangled, and a period that ends an abbreviation is not treated as a sentence boundary.

2. **Smart spacing at the seams.** Today a single space is always tacked onto the end of every dictated sentence, no matter the context. This change drops that blanket trailing space and instead adds or omits a space on *each* side based on what is actually there — words and numbers get a space, opening/closing punctuation does not — and double spaces are never produced.

3. **Punctuation cleanup.** Removes punctuation that clashes with what follows: a period before a closing `)`/`,`/`?`, a duplicated terminal mark (`!!`, `,,`), a mid-sentence period before lowercase text, a stray ellipsis dropped into the middle of a sentence, and — when you are inserting right before an existing comma — any terminal mark the model appended (`,` `.` `;` `:` `!` `?` `…`), so the comma already in the document stays the single separator (abbreviation periods are preserved).

4. **Correct replacement of selected text.** When you dictate over a selection (ordinary dictation, not the AI "edit selected text" command mode), the spaces that belonged to the selection are restored, so the replacement does not glue itself to the neighboring words. The selected text is read untrimmed so those edge spaces survive.

5. **Custom-vocabulary casing.** Any word you defined (a name, a brand, an acronym) is restored to your exact spelling wherever it appears in the dictation — not just the first word — matched as a whole word so it never corrupts substrings (`MacOS` stays `MacOS`).

6. **Reliable, non-destructive context reading in web / Electron apps.** Uses the WebKit/Chromium text-position API (TextMarkers) to read the real text around the cursor in apps where the plain integer cursor offset lies (Chromium reports a fraudulent `(0,0)`), with a tree search, the non-destructive start-of-recording snapshot, and — only as a genuine last resort, only in web/Electron — a copy/paste read. The accessibility tree is woken just-in-time so the very first dictation in an app still works. Reading context never moves the caret.

7. **The cleanup model sees a bounded slice of the surrounding text.** The post-processing model now also receives the text immediately before and after the cursor as a reference — so it can match the spelling, casing, and tone of what is already there (for example, a recipient's name visible just before the cursor). That slice is **bounded** (a couple of sentences before, one after, capped by length, flattened to one line) and the prompt explicitly instructs the model never to copy the surrounding text into its output — together this prevents the model from "restoring" already-typed text.

8. **Recognizes real paragraph / page breaks.** Line, paragraph, and page breaks are normalized to a canonical newline before the formatter looks at them, so capitalization after a true paragraph break works regardless of how the app encodes the break (CR, CRLF, Unicode line/paragraph separators, next-line, or form-feed).

9. **Full pipeline visibility in the Run Log.** Each dictation is now traced end to end — what was read around the cursor (and by which method), what the cleanup model received and returned, what the new deterministic formatting stage did (in plain English), and the final inserted text — plus a per-session metric of how often the precise read paths win versus the copy/paste fallback.

---

## How it works (the short version)

- Reads up to a few hundred characters before and after the cursor through the macOS Accessibility API, **without ever typing, selecting, or moving the cursor in 99% of the time** on the common paths.
- Native controls use the standard API. In web / Electron views, a real form field (`<textarea>`/`<input>`) is read natively too — it exposes the same self-scoped value+caret as a native field — while rich contenteditable editors use the WebKit/Chromium TextMarker attributes, because their integer cursor position is unreliable. A breadth-first search for the web area, a non-destructive snapshot taken when recording started, and — last of all, web/Electron only — a copy/paste read are graceful fallbacks.
- A small, bounded slice of that context goes to the cleanup model as a hint (never copied into the output); the full context feeds the deterministic formatter.
- The deterministic formatter then runs four ordered phases — normalize, punctuation, capitalization, spacing — plus the vocabulary-casing pass.
- None of this changes *what* you said; it only changes how the already-cleaned text is fitted into place.

---

### 1. Reading the surrounding text — an explicit extraction ladder

Context is read by `AccessibilityTextReader`, which tries methods in priority order and stops at the first that returns usable text. The ladder is deliberately ordered **non-destructive first, the one destructive method last**, and it branches on whether the focused element is a native control or a web/Electron view (detected by the presence of `AXDOM*`/`AXWeb*` attribute names):

- **Native (AppKit):** read `kAXValue` + `kAXSelectedTextRange` and slice around the selection. Reliable; nothing else is needed; the destructive fallback is never reached for native apps.

- **Web / Electron:**
  1. **Native form-control read** — a genuine HTML `<textarea>`/`<input>` is exposed by Safari/Chromium as an accessibility *leaf* (`AXTextField`/`AXTextArea`) whose text is its own self-scoped `kAXValue` + `kAXSelectedTextRange` — the *same contract* a native macOS field uses. These are read with the native offset path **first**, because a `<textarea>`'s text is **not** part of the surrounding page's TextMarker document: a TextMarker read would climb to the `AXWebArea` and wrongly capture unrelated page content sitting next to the control (e.g. a live-region "response ready" status just above a chat input) as the preceding text. The native value is field-scoped, so it can never leak neighbouring page text. The gate is site-agnostic — it keys on the element's accessibility contract (plain leaf text control + honest caret), never on per-website markup — and bails to TextMarkers for anything that is not a plain control or that reports the fraudulent `(0,0)` cursor.
  2. **TextMarkers** — the WebKit/Chromium `AXTextMarker*` parameterized attributes, with the underlying C functions (`AXTextMarkerRangeCopyStartMarker` / `…EndMarker` / `AXTextMarkerRangeCreate`) resolved at runtime via `dlsym`. The tool for **rich contenteditable** editors (where there is no single honest value). TextMarkers address positions by opaque markers instead of integer offsets, so they are immune to the fraudulent `(0,0)` cursor Chromium reports through `kAXSelectedTextRange`; the read is scoped to the field (or, if field-scoping is unavailable, clamped to the caret's current line) so it does not reach back into page content.
  3. **Offset AX** with an explicit fraudulent-`(0,0)` guard.
  4. **Breadth-first search** for the focused element inside an `AXWebArea` (when the app hands us the wrong element from its root).
  5. **Start-of-recording snapshot** — a non-destructive read taken when recording began (the text before the cursor does not change while you dictate, so it is still valid at paste time).
  6. **Keyboard copy/paste read** — `Cmd+Shift+Up/Down` to select, `Cmd+C` to read the clipboard, then collapse — the only method that moves the caret. It runs **only** when every method above failed, **only** in web/Electron, and **never** on the time-sensitive release path before the audio uploads. The original clipboard is saved and restored.
  7. **Blind** — return no context; the formatter degrades gracefully (it preserves the model's own casing rather than forcing a sentence start).

The accessibility tree is woken just-in-time by setting the private `AXManualAccessibility` attribute (Chromium keeps its tree asleep for performance, and builds it lazily). This replaced an older trick that nudged the tree awake by synthesizing arrow keys — which is exactly what could shift the caret — so that key-based wake and its dead parameters were removed.

**Robustness when the machine is busy.** Reading another app's text over the Accessibility API is a cross-process request, so a CPU-starved or unresponsive target app could otherwise stall the read. Three safeguards bound that: (a) a **per-request timeout** (`AXUIElementSetMessagingTimeout`, 2 s) so a single Accessibility call can't block for the ~6 s system default — a slow app degrades to a fallback instead of hanging; (b) the heavy read runs **off the main thread**, so the app's own UI never freezes while waiting; and (c) the just-in-time Accessibility wake is **kept alive until the paste (and its re-read) finishes**, instead of on a fixed timer, so a slow paste can't have the tree torn down underneath it. These bound *latency*, not success: a non-empty read is never guaranteed for any cross-process Accessibility read, but the pipeline always completes and the dictation is never dropped — at worst it falls back to a plain insertion.

### 2. Bounding the context sent to the cleanup model (`SurroundingTextLimiter`)

Feeding the model the full surrounding text invites it to echo that text back into its answer (duplicating already-typed content). A new, self-contained `SurroundingTextLimiter` produces a small, bounded slice for the prompt only:

- keep at most the **last 2 sentences** before the cursor and the **first sentence** after it;
- enforce a **character budget** with a tolerance band, reducing the sentence count before truncating;
- if a single sentence is still over budget, **truncate at a word boundary** (never mid-word; an over-long single token is kept whole rather than split);
- **flatten** whitespace to single spaces for the prompt.

This affects the prompt only — **the deterministic formatter still receives the full, unflattened text with its real line breaks.** The user-facing prompt was also strengthened: the surrounding text is framed as reference-only, with a positive role description plus an explicit "never copy, repeat, continue, or complete the surrounding text" instruction.

**Prompt hygiene — no scaffolding tokens leak into the output (two layers):** the prompt labels its sections with uppercase tokens (`PRECEDING_TEXT`, `FOLLOWING_TEXT`, …). A model occasionally *narrates* an empty labeled section and leaks the token into the cleaned text (observed: a stray line "FOLLOWING_TEXT is not relevant…" when the cursor was at the end). Fixed at the root by **omitting an empty section from the prompt entirely** (no blank labeled field to comment on), plus a defensive backstop (`PromptScaffoldingSanitizer`) that **strips any output line beginning with a scaffolding token** — inline mentions and lone-token content are preserved, so legitimate dictation is never harmed.

### 3. The deterministic formatter (`ContextualFormattingService`), four ordered phases

1. **Normalize** — trim, standardize ellipses, and canonicalize every line/paragraph/page break (`\r`, `\r\n`, U+2028, U+2029, U+0085, U+000C) to `\n` so the later phases recognize a real break regardless of encoding.
2. **Punctuation** — drop a stray mid-sentence ellipsis; collapse a duplicated terminal mark; remove a trailing period that collides with a following closing/terminal mark or with lowercase continuation (abbreviations exempt); and, before an existing comma, drop any terminal mark the model appended.
3. **Capitalization** — decide the first letter from the preceding context (field start, after `.?!…`, after a line/paragraph break, after a bullet, after a sentence-opening quote/parenthesis → uppercase; mid-sentence → lowercase), while preserving abbreviations, the pronoun `I`/contractions, and all-caps acronyms, and treating an abbreviation-ending period as *not* a boundary.
4. **Spacing** — add/omit a leading and trailing space from what is actually on each side, restore edge spaces consumed when replacing a selection, and collapse any accidental double space.

The cursor-position semantics were also tightened: when only one side of the cursor can be read, the position is reported as `middle`/`unknown` rather than incorrectly claiming `end`.

### 4. Faithful diagnostics

`ContextDebugLabels` is the single source of labels so the live debug panel and the saved history can never disagree, and `ContextReadMetrics` tallies per session how often each read method won. The recorded extraction method always reflects the rung that actually produced the context.

### 5. Merge-safety / additivity

The shared backbone files are touched **additively** so this branch stays merge-compatible with other in-flight work regardless of merge order: the cleanup-prompt change is new defaulted parameters (the existing call sites and contested lines stay byte-identical); the surrounding-text capture lives in the feature's own region of `collectContext`; and the heavy logic lives in new, feature-owned files. The two largest files (`AccessibilityTextReader`, `ContextualFormattingService`) and `SurroundingTextLimiter` exist only on this branch.

---

### New

| File | What it adds |
|------|--------------|
| `ContextualFormattingService.swift` | The deterministic formatter: four ordered phases (normalize, punctuation, capitalization, spacing) plus vocabulary-casing, and line-break normalization. The brain of the feature. |
| `AccessibilityTextReader.swift` | Reads the text around the cursor via the extraction ladder: native API (including real web `<textarea>`/`<input>` form fields, read field-scoped to avoid leaking page text), WebKit/Chromium TextMarkers for rich contenteditable (resolved with `dlsym`), web-area tree search, and the last-resort copy/paste fallback. Non-destructive on the common paths. |
| `SurroundingTextLimiter.swift` | Bounds the surrounding text sent to the cleanup model (sentence cap + character budget with tolerance + word-safe truncation + whitespace flatten) so the model can use it as a hint without echoing it back. |
| `PromptScaffoldingSanitizer.swift` | Defensive backstop that strips any output line beginning with a prompt-scaffolding token (`PRECEDING_TEXT`/`FOLLOWING_TEXT`/…) the model may have leaked, while preserving inline mentions and lone-token content. |
| `AccessibilityWakeManager.swift` | Wakes the accessibility tree just-in-time via `AXManualAccessibility` (and observes focus/selection) so the first read in an app is reliable — without synthesizing any keystrokes. |
| `ContextDebugLabels.swift` | Single source of plain-English labels for the debug panels, so the live and saved views can never disagree. |
| `ContextReadMetrics.swift` | Session tally of how the surrounding text was read (precise vs fallback), logged per dictation. |

### Modified

| File | What changed and why |
|------|----------------------|
| `AppState.swift` | Wires the formatter into the paste pipeline; captures context just-in-time at the end of dictation; orders the fallbacks so the non-destructive start-of-recording snapshot is preferred and the keyboard read is a true last resort; normalizes the resolved context's line breaks once; passes the custom vocabulary; logs the read method/metric; and only fires the selection-clearing keystroke when a live selection actually exists. |
| `AppContextService.swift` | Collects the surrounding text around the cursor (non-destructively) alongside the rest of the context, and reads the selected text untrimmed for correct space restoration. Screenshot and activity-summary behavior left identical to upstream. |
| `PostProcessingService.swift` | Passes a bounded slice of the surrounding text into the cleanup prompt as a reference (via `SurroundingTextLimiter`), with a strengthened never-copy instruction — added through new defaulted parameters so the change is purely additive and stays merge-compatible with other work on this file. |
| `PipelineHistoryItem.swift` | New optional fields (surrounding text, cursor position, applied rule, read method, app kind) so history entries are backward-compatible. |
| `PipelineHistoryStore.swift` | Lightweight Core Data migration that adds the new fields to old stores without data loss. |
| `PipelineDebugContentView.swift` | The live debug panel shows the smart-paste result in plain English. |
| `PipelineDebugPanelView.swift` | Passes the new context fields through to the panel. |
| `SettingsView.swift` | The saved-history view shows the same plain-English smart-paste summary. |

---

## Notes

- **Never disturbs the document.** Reading context does not type, select, or move the cursor on the common paths; the one method that does (the web/Electron copy/paste read) is a guarded last resort that restores the clipboard.
- **Backward compatible.** New history fields are optional and the Core Data store migrates in place, so existing history keeps working.
- **Graceful degradation.** Every read path is guarded; if an app exposes nothing, the feature backs off (blind) and the dictation behaves exactly as before.
- **Self-contained.** No third-party libraries are added.
- **Additive & merge-safe.** Backbone files change only additively (defaulted parameters, feature-owned regions); the heavy logic is in new files.
- **As little "magic" as the platform allows.** macOS exposes no public API for the text around the cursor in web/Electron views, so the unavoidable bits — the private `AXManualAccessibility` wake and the WebKit/Chromium TextMarker SPIs — are isolated to clearly-labeled rungs of one explicit extraction ladder, documented, bounded by a per-request timeout, and used **only** when the clean native path doesn't apply. The deterministic formatter is plain, testable string logic with no platform tricks, and every path degrades gracefully. The goal was exactly the "clean implementation without too much voodoo magic" raised in #200: the cleverness is fenced off, layered, and reversible rather than spread through the codebase.

## Summary by CodeRabbit

* **New Features**
  * Added richer context capture for dictation, including nearby text, cursor position, app type, and formatting details.
  * Added a Smart Paste formatting view and copyable debug log for easier troubleshooting.
  * Improved context-aware post-processing so inserted text can adapt to surrounding text more accurately.

* **Bug Fixes**
  * Made app text reading more reliable across native apps, web apps, and Electron apps.
  * Improved paste behavior and reduced chances of losing spacing, punctuation, or selection state.
  * Fixed cases where captured text could be incomplete or trimmed incorrectly.

### DIFF
--- a/Sources/AccessibilityTextReader.swift
+++ b/Sources/AccessibilityTextReader.swift
@@ -1,0 +1,1123 @@
+import Foundation
+import ApplicationServices
+import AppKit
+import OSLog
+import Darwin   // dlsym, for resolving private AXTextMarker C functions at runtime
+
+private let axLog = Logger(subsystem: "com.zachlatta.freeflow", category: "AccessibilityText")
+
+// MARK: - SurroundingTextSnapshot
+
+/// Text context captured around the cursor from the focused text field.
+/// All fields are optional; nil means the app does not expose accessible text.
+struct SurroundingTextSnapshot: Sendable {
+    /// Up to N characters before the cursor (or selection start).
+    let precedingText: String?
+    /// Up to N characters after the selection end (never includes selected text).
+    let followingText: String?
+    /// Semantic position of the cursor within the field.
+    let cursorPosition: CursorPosition
+    /// Which extraction technique successfully read the surrounding text.
+    let extractionMethod: ExtractionMethod
+    /// What kind of app the text came from. Diagnostics/UX only — never drives logic.
+    var appKind: AppKind = .unknown
+
+    /// Semantic position of the cursor within the focused field.
+    enum CursorPosition: String {
+        case start    // index 0
+        case middle   // somewhere in the body
+        case end      // at or past the last character
+        case empty    // field has no text
+        case unknown  // AX API did not expose position info
+    }
+
+    /// Which Accessibility technique successfully read the surrounding text.
+    enum ExtractionMethod: String {
+        case axAPI               // Native kAXValueAttribute worked
+        case axWebTextMarker     // WebKit/Chromium TextMarker parameterized attributes
+        case axWebAreaBFS        // BFS located the focused element inside an AXWebArea
+        case keyboardSimulation  // Cmd+Shift+Up/Down clipboard-selection fallback used
+        case unknown             // No technique could get context
+    }
+
+    /// The kind of application hosting the focused text field.
+    /// Used only for human-readable diagnostics — it never changes how text is read.
+    enum AppKind: String {
+        case native   // A standard Mac app (its text fields speak the native accessibility API)
+        case webView  // An Electron / Chromium / Safari-style app (text lives inside a web page)
+        case unknown  // We could not tell (there was no focused field to inspect)
+    }
+
+    /// True when at least one side of the cursor is readable.
+    var hasContext: Bool { precedingText != nil || followingText != nil }
+
+    /// Returns a copy of this snapshot tagged with the given app kind.
+    func withAppKind(_ kind: AppKind) -> SurroundingTextSnapshot {
+        var copy = self
+        copy.appKind = kind
+        return copy
+    }
+}
+
+// MARK: - AccessibilityTextReader
+
+/// Reads text surrounding the cursor via the macOS Accessibility API.
+///
+/// Covers:
+/// - Native AppKit controls (NSTextView, NSTextField, etc.)
+/// - Electron / Chromium web views (e.g. editor and chat apps)
+/// - Browser text fields (Safari, Chrome, Firefox)
+/// - Qt and Java/Swing apps via their respective AX bridges
+/// - Terminal emulators — limited, field-level only
+///
+/// Falls back gracefully: returns nil values for unsupported apps.
+enum AccessibilityTextReader {
+
+    /// Per-message AX IPC timeout (seconds). Bounds each cross-process Accessibility request so a
+    /// CPU-starved target app degrades to a fallback instead of blocking for the ~6s system default.
+    private static let axMessagingTimeoutSeconds: Float = 2.0
+
+    // MARK: Public API
+
+    /// Synchronous read — works for native AppKit controls that expose kAXValueAttribute.
+    /// For Electron/web views (where AX may not be awake), this returns nil for surrounding text;
+    /// the async collectContext() path handles those via BFS and keyboard fallback.
+    static func readSurroundingText(
+        from appElement: AXUIElement,
+        maxBefore: Int = 300,
+        maxAfter: Int = 400
+    ) -> SurroundingTextSnapshot {
+        // Cap per-message AX IPC so a CPU-starved target app can't block this read for the
+        // ~6s system default; a slow app then degrades to a fallback instead of stalling.
+        _ = AXUIElementSetMessagingTimeout(appElement, axMessagingTimeoutSeconds)
+        guard let focused = focusedElement(in: appElement) else {
+            axLog.info("readSurroundingText(sync): no focused element found")
+            return .init(precedingText: nil, followingText: nil, cursorPosition: .unknown, extractionMethod: .unknown)
+        }
+
+        // Probe app kind once; we have a focused element to inspect from here on.
+        let kind = detectAppKind(of: focused)
+
+        // Genuine web form controls (<textarea>/<input>) expose an honest, self-scoped
+        // kAXValue + caret — read them natively FIRST so a TextMarker read can't capture
+        // surrounding page text as context (see nativeReadForWebFormControl).
+        if kind == .webView,
+           let nativeSnap = nativeReadForWebFormControl(focused, maxBefore: maxBefore, maxAfter: maxAfter) {
+            axLog.info("readSurroundingText(sync): done via native form-control read — prec=\(nativeSnap.precedingText?.count ?? 0)ch foll=\(nativeSnap.followingText?.count ?? 0)ch")
+            return nativeSnap.withAppKind(.webView)
+        }
+
+        // Web/Electron: prefer TextMarkers — immune to the fraudulent (0,0) cursor that
+        // kAXSelectedTextRange returns in Chromium. Falls through to the native path on nil.
+        if kind == .webView,
+           let markerSnap = extractViaTextMarkers(focused, maxBefore: maxBefore, maxAfter: maxAfter),
+           markerSnap.hasContext {
+            axLog.info("readSurroundingText(sync): done via TextMarkers — prec=\(markerSnap.precedingText?.count ?? 0)ch foll=\(markerSnap.followingText?.count ?? 0)ch")
+            return markerSnap.withAppKind(.webView)
+        }
+
+        // Read full text — may fail in Electron where the AX tree is not yet awake.
+        var valueRef: CFTypeRef?
+        guard AXUIElementCopyAttributeValue(focused, kAXValueAttribute as CFString, &valueRef) == .success,
+              let fullText = valueRef as? String else {
+            axLog.info("readSurroundingText(sync): kAXValueAttribute unavailable — async read required")
+            return .init(precedingText: nil, followingText: nil, cursorPosition: .unknown, extractionMethod: .unknown, appKind: kind)
+        }
+
+        guard !fullText.isEmpty else {
+            return .init(precedingText: "", followingText: "", cursorPosition: .empty, extractionMethod: .axAPI, appKind: kind)
+        }
+
+        let nsText = fullText as NSString
+        let totalLength = nsText.length
+
+        // Read cursor range — synchronous, no keyboard simulation possible here.
+        var rangeRef: CFTypeRef?
+        guard AXUIElementCopyAttributeValue(focused, kAXSelectedTextRangeAttribute as CFString, &rangeRef) == .success,
+              let rangeVal = rangeRef,
+              CFGetTypeID(rangeVal) == AXValueGetTypeID() else {
+            axLog.info("readSurroundingText(sync): no range info — returning tail as preceding")
+            let tail = nsText.substring(from: max(0, totalLength - maxBefore))
+            return .init(precedingText: tail, followingText: "", cursorPosition: .unknown, extractionMethod: .axAPI, appKind: kind)
+        }
+
+        var cfRange = CFRange(location: 0, length: 0)
+        guard AXValueGetValue(unsafeBitCast(rangeVal, to: AXValue.self), .cfRange, &cfRange) else {
+            return .init(precedingText: nil, followingText: nil, cursorPosition: .unknown, extractionMethod: .unknown, appKind: kind)
+        }
+
+        let selStart = min(max(cfRange.location, 0), totalLength)
+        let selEnd   = min(selStart + max(cfRange.length, 0), totalLength)
+
+        let beforeStart    = max(0, selStart - maxBefore)
+        let precedingSlice = nsText.substring(with: NSRange(location: beforeStart, length: selStart - beforeStart))
+        let afterLength    = min(maxAfter, totalLength - selEnd)
+        let followingSlice = afterLength > 0
+            ? nsText.substring(with: NSRange(location: selEnd, length: afterLength))
+            : ""
+
+        let position: SurroundingTextSnapshot.CursorPosition
+        if selStart == 0               { position = .start }
+        else if selStart >= totalLength { position = .end }
+        else                           { position = .middle }
+
+        axLog.info("readSurroundingText(sync): prec=\(selStart - beforeStart)ch foll=\(afterLength)ch pos=\(position.rawValue, privacy: .public) kind=\(kind.rawValue, privacy: .public)")
+        return .init(precedingText: precedingSlice, followingText: followingSlice, cursorPosition: position, extractionMethod: .axAPI, appKind: kind)
+    }
+
+    /// Async read that auto-syncs the AX tree for Electron/web views before reading.
+    /// Always prefer this variant inside async contexts (e.g., collectContext).
+    static func readSurroundingTextWithSync(
+        from appElement: AXUIElement,
+        maxBefore: Int = 300,
+        maxAfter: Int = 400,
+        allowKeyboardSelection: Bool = false
+    ) async -> SurroundingTextSnapshot {
+        let readStart = ContinuousClock.now
+        // Cap per-message AX IPC (see readSurroundingText) so a busy target app can't block ~6s/call.
+        _ = AXUIElementSetMessagingTimeout(appElement, axMessagingTimeoutSeconds)
+
+        // ── Attempt 1: Standard focused element ──
+        var focused = focusedElement(in: appElement)
+        var didBFS = false
+
+        // ── Attempt 2: BFS AXWebArea fallback ──
+        // Standard AX focused-element query often returns nil for Electron/Chromium apps
+        // because the text element lives inside a web content subtree, not at the app root.
+        if focused == nil {
+            axLog.info("readSurroundingTextWithSync: standard focused element nil — starting BFS")
+            focused = searchForWebAreaFocusedElement(in: appElement)
+            didBFS = true
+            if focused == nil {
+                axLog.warning("readSurroundingTextWithSync: BFS found no AXWebArea — returning empty snapshot")
+                return .init(precedingText: nil, followingText: nil, cursorPosition: .unknown, extractionMethod: .unknown)
+            }
+        }
+
+        guard let focused else {
+            return .init(precedingText: nil, followingText: nil, cursorPosition: .unknown, extractionMethod: .unknown)
+        }
+
+        // Web/Electron detection is invariant for `focused` within one read — compute the AX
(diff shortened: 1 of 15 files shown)


pianoandglass:
Reworking the Code for an even cleaner implementation, removing obsolete steps and mechanisms.

PR #248, opened 2026-07-01
Add Preserve exact wording toggle to skip LLM cleanup

ChaoticQubit:
## Summary

Add a **Preserve exact wording** toggle in Settings → Cleanup that, when enabled, skips the LLM post-processing step and pastes the raw transcript from the transcription service verbatim, including profanity, filler words, and informal wording.

## Motivation

There are legitimate cases where users want the raw transcript to reach the target text field unchanged:

- **Verbatim quotes and interviews** — journalism, research notes, court transcripts, etc., where any silent cleanup by the model could alter meaning.
- **Filler-word coaching** — some users record dictation deliberately to review how often they say "um", "like", etc., which the LLM currently strips.
- **Explicit language** — the cleanup model may sanitize wording the user actually intended.
- **Offline / rate-limited use** — post-processing is the slowest and most failure-prone part of the pipeline; letting users opt out gives them a fast, deterministic path that only depends on the transcription service.

## Behavior

- Default: **off**. No behavior change for existing users.
- When on: `processTranscript` returns the trimmed raw transcript early with a new `.preservedExactWording` `TranscriptProcessingOutcome` case. Post-processing is skipped entirely.
- **Voice macros and Edit Mode still run.** The early-return sits after the voice-macro lookup and does not affect the `commandInvocation` branches — those keep their existing semantics.
- The new outcome surfaces the status "Preserved exact wording, skipped post-processing" in the debug status line, keeping the existing status-message contract.

## Changes

- `Sources/AppState.swift` (+34, −1):
  - Storage key `preserveExactWordingStorageKey`.
  - `@Published var preserveExactWording: Bool` with `didSet` persistence, mirroring the existing `preserveClipboard` pattern.
  - Loaded in `init` from `UserDefaults` (defaults to `false`).
  - New `TranscriptProcessingOutcome.preservedExactWording` case with `statusMessage()`.
  - Early-return guard in `processTranscript` after the voice-macro branch and before `postProcessingService.postProcess(...)`.
- `Sources/SettingsView.swift` (+15):
  - New `SettingsCard("Cleanup", icon: "sparkles")` placed between "Edit Mode" and "Clipboard" in `GeneralSettingsView`.
  - `cleanupSection` with a `Toggle` bound to `appState.preserveExactWording` and a caption explaining the behavior.

Total diff: **+50 / −1** across 2 files, no new dependencies.

## Test plan

- Toggle off (default): identical behavior to `main`. Verified against the existing `postProcessingSucceeded` / `postProcessingFailedFallback` flows.
- Toggle on: raw transcript from Groq is pasted with no LLM pass. Voice macros still fire (verified with a test macro). Edit Mode still routes through `commandModeSucceeded`.
- Setting persists across app restarts.
- Status line reflects the new outcome.

## Notes for maintainers

Happy to adjust the label ("Preserve exact wording") or card placement if you have a preferred wording / grouping. The Cleanup card is new; if you'd rather fold the toggle into an existing card (e.g. "Edit Mode" or a new "Post-processing" grouping), the change is one-line.

## Summary by CodeRabbit

* **New Features**
  * Added a new **“Preserve exact wording”** setting in General settings under a new **Cleanup** section.
* **Behavior Changes**
  * When enabled, transcript cleanup is skipped so the original wording is preserved.
  * If an Output Language is set, translation runs in a literal “verbatim” style and preserved-text outcomes are reflected in status.
  * Updated clipboard/raw-dictation fallback handling to treat translation-failed preserved wording consistently.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -1191,7 +1191,8 @@ final class AppState: ObservableObject, @unchecked Sendable {
                     postProcessingService: postProcessingService,
                     customVocabulary: capturedCustomVocabulary,
                     customSystemPrompt: capturedCustomSystemPrompt,
-                    outputLanguage: self.outputLanguage
+                    outputLanguage: self.outputLanguage,
+                    preserveExactWording: self.preserveExactWording
                 )
                 finalTranscript = result.finalTranscript
                 processingStatus = Self.statusMessage(
@@ -2436,6 +2437,8 @@ final class AppState: ObservableObject, @unchecked Sendable {
         case postProcessingSucceeded
         case postProcessingFailedFallback
         case preservedExactWording
+        case preservedExactWordingTranslated
+        case preservedExactWordingTranslationFailedFallback
         case commandModeSucceeded(invocation: CommandInvocation)
         case commandModeFailedFallback(invocation: CommandInvocation)
 
@@ -2453,6 +2456,10 @@ final class AppState: ObservableObject, @unchecked Sendable {
                     : "Post-processing failed, using raw transcript"
             case .preservedExactWording:
                 return "Preserved exact wording, skipped post-processing"
+            case .preservedExactWordingTranslated:
+                return "Preserved exact wording, translated to output language"
+            case .preservedExactWordingTranslationFailedFallback:
+                return "Verbatim translation failed, using untranslated raw transcript"
             case .commandModeSucceeded(let invocation):
                 return "Edit mode succeeded (\(invocation.rawValue))"
             case .commandModeFailedFallback(let invocation):
@@ -2468,7 +2475,8 @@ final class AppState: ObservableObject, @unchecked Sendable {
         postProcessingService: PostProcessingService,
         customVocabulary: String,
         customSystemPrompt: String,
-        outputLanguage: String = ""
+        outputLanguage: String = "",
+        preserveExactWording: Bool
     ) async -> (finalTranscript: String, outcome: TranscriptProcessingOutcome, prompt: String) {
         let trimmedRawTranscript = rawTranscript.trimmingCharacters(in: .whitespacesAndNewlines)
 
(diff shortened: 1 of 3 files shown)


ChaoticQubit:
Thanks for the review — CodeRabbit's callout about the `outputLanguage` interaction was the important one. I agree the previous behaviour was a bug, not just a documentation gap: silently dropping translation when the user had configured an Output Language defeats their setting.

I've pushed a follow-up commit (`8224655`) that fixes the underlying behaviour rather than only the copy:

**New routing in `processTranscript`**

- `preserveExactWording=off` — unchanged (regular `postProcess`).
- `preserveExactWording=on`, no Output Language — skip the LLM entirely, as before.
- `preserveExactWording=on`, Output Language set — route through a new **verbatim translation** path.

**New `PostProcessingService.translateVerbatim`**

Shares the primary/fallback model selection and timeout of the regular cleanup path, but uses a stripped-down translate-only prompt:

- Translate word-for-word into the target language.
- Preserve every filler word ("um", "uh", "like"), false start, and repetition — translate them into the closest natural equivalent rather than deleting them.
- No rewording, restructuring, grammar correction, or reformatting.
- No punctuation additions beyond what the target language grammatically requires.
- Keep profanity, slang, and explicit language intact.
- Output only the translated text (no wrappers, no explanations).

**Also addressed:**

- Threaded `preserveExactWording` through as an explicit function parameter (finding #2), matching how `outputLanguage`, `customVocabulary`, and `customSystemPrompt` are already passed.
- Two new `TranscriptProcessingOutcome` cases (`preservedExactWordingTranslated`, `preservedExactWordingTranslationFailedFallback`) with status messages.
- `preservedExactWordingTranslationFailedFallback` participates in `shouldPersistRawDictationFallback` so users can retry the failing session, matching the semantics of `postProcessingFailedFallback`.
- Rewrote the Settings copy to describe the new, correct behaviour ("If Output Language is set, the transcript is still translated, but the translation is literal…").

Total diff after the follow-up: **+236 / −9** across 3 files.

Happy to squash the two commits if you prefer a single commit in the history.

ChaoticQubit:
Good catch — that one is real. `sanitizePostProcessedTranscript` treats the literal string `"EMPTY"` as a sentinel because the cleanup system prompt explicitly tells the model to return that when there's nothing to paste. The verbatim translation prompt has no such instruction, so reusing the same sanitizer risks silently dropping a legitimate literal translation of the word "empty" into the target language.

Fixed in `2003af9`. Added `sanitizeVerbatimTranslation` which only trims whitespace and strips wrapping quotes — no `"EMPTY"` sentinel handling.

PR #252, opened 2026-07-03
feat: background HTTP pre-transcription while recording

JaPossert:
## Summary

- Adds a new **"Begin transcribing while recording"** setting that submits audio to the transcription provider in **28-second background chunks** while the user speaks
- By the time the user presses stop, the majority of the recording is already processed — only the trailing segment (< 28 s) needs transcribing on stop, dramatically reducing perceived latency for long dictations
- Works with **any standard HTTP transcription endpoint** (Groq, local Whisper, faster-whisper, etc.) — no WebSocket or special server required, unlike the existing realtime mode
- Each completed chunk's transcript is **appended to a durable `.txt` file** in the audio directory immediately after transcription — an app crash loses at most one 28-second chunk
- The setting **defaults to enabled** on first launch (same nil-check pattern as the Press Enter voice command), so users benefit immediately without manual configuration

### How it works

A new `PrefetchTranscriber` class accumulates the raw PCM16 audio stream (`AudioRecorder.onPCM16Samples`, 24 kHz mono). Every 28 seconds it writes a temporary WAV file and submits it to the configured `TranscriptionService` endpoint in a serial background task chain. When recording stops, `commitAndAwaitFinal()` waits for any in-flight chunk, transcribes the remaining tail, and returns the joined transcript. If the merged result is empty (network error, etc.) it falls back to the regular full-file upload path.

Each chunk's transcript is appended via `FileHandle` to a timestamped `.txt` file in `~/Library/Application Support/FreeFlow Dev/audio/`. In a 3-hour session the transcriber holds at most ~2 × 1.35 MB of PCM16 in RAM at any time (one chunk being transcribed + partial next chunk buffer).

The existing **realtime WebSocket mode** (`realtimeStreamingEnabled`) takes priority when both settings are enabled, so the two paths are non-conflicting.

### Files changed

| File | Change |
|------|--------|
| `PrefetchTranscriber.swift` | New class: PCM16 accumulation, 28-second chunking, serial task chain, durable file append |
| `AppState.swift` | New `prefetchTranscriptionEnabled` property (default: on); recording start/stop/cancel wiring; durable save-URL generation |
| `SettingsView.swift` | New "Begin transcribing while recording" toggle above existing realtime toggle |

Relates to #250 (URLSession timeout fixes for local LLM inference).

## Summary by CodeRabbit

* **New Features**
  * Added an option to transcribe in background while recording by prefetching audio chunks so results are ready sooner.
* **Bug Fixes**
  * Improved transcript selection on stop: realtime transcription is preferred, with background prefetch used as fallback before file-based transcription.
  * Ensured background prefetch processing is correctly started, torn down, and discarded when realtime is disabled or recording ends.
* **Documentation**
  * Updated the settings text to clarify background transcription behavior and that realtime takes priority when both are enabled.

### DIFF
--- a/Sources/AppState.swift
+++ b/Sources/AppState.swift
@@ -694,7 +694,9 @@ final class AppState: ObservableObject, @unchecked Sendable {
         let keepDictationInClipboardHistory = UserDefaults.standard.bool(forKey: keepDictationInClipboardHistoryStorageKey)
         let realtimeStreamingEnabled = UserDefaults.standard.bool(forKey: realtimeStreamingEnabledStorageKey)
         let realtimeStreamingModel = UserDefaults.standard.string(forKey: realtimeStreamingModelStorageKey) ?? ""
-        let prefetchTranscriptionEnabled = UserDefaults.standard.bool(forKey: prefetchTranscriptionEnabledStorageKey)
+        let prefetchTranscriptionEnabled = UserDefaults.standard.object(forKey: prefetchTranscriptionEnabledStorageKey) == nil
+            ? true
+            : UserDefaults.standard.bool(forKey: prefetchTranscriptionEnabledStorageKey)
         let dictationAudioInterruptionEnabled = UserDefaults.standard.bool(
             forKey: dictationAudioInterruptionEnabledStorageKey
         )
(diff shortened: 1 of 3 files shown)


JaPossert:
Closing this one too, for the same reason: I would rather contribute findings and a branch you can read than a pull request written with AI assistance. The implementation remains on my fork if it is useful.

Issue #212, opened 2026-05-26
Configurable/floating recording indicator


Clleo:
Hi Zach,

First of all, thank you for creating Freeflow.
It’s one of the most comfortable and thoughtfully designed voice-input tools I’ve used.

I’d like to share a small optional UI idea that might improve usability for some users.

Currently, the voice input indicator is displayed at the top of the screen, and overall I actually like this design choice — it looks clean and unobtrusive. However, when the internet connection is slow or unstable, I often have to constantly look upward to check whether speech recognition is still active or if it has frozen.

It might be useful to add an optional setting that allows users to choose where the voice input indicator appears:
— either at the top, as it is now,
— or near the actual text input area, close to the cursor/text field.

I think this could be more comfortable for some users and would reduce the need to constantly shift attention across the screen while dictating.

marcbodea:
Sounds like a good idea, let's see how easy the implementation is

mvanhorn:
Opened #242 for this. It adds a "Show at" setting (Top of screen, the current default, or Near cursor). Near-cursor anchors the drop-down pill just below the active text caret (via Accessibility, falling back to the pointer), clamped to the visible screen so it never clips at an edge. The minimalist notch layout stays top-anchored, and the default behavior is unchanged.

PR #254, opened 2026-07-05
Timeout fix, Edit Mode in Electron apps, cursor-aware cleanup, Prompt Mode, vocabulary mappings, custom sounds, faster stop-to-paste

### LINKED ISSUE #107: Allow custom sounds (or more options) for recording start, stop, and error feedback
Small feature request on the audio feedback side.

The three built-in sounds (Tink on start, Pop on stop, Basso on error) work fine, but I have gotten used to a different set of chimes from another dictation app and my muscle memory keeps reaching for those. More options around the feedback sounds would be great — either a wider catalog of built-in choices, or the ability to point FreeFlow at my own audio files.

## Suggestion

In Settings, next to each of the three feedback events, something like:

- A dropdown with more built-in options
- And/or a "Choose file..." option that opens a file picker for an audio file on disk (`.aiff`, `.wav`, `.mp3`)
- Preview button next to each (see also #106 for the muted-system alert on that preview)

Would love to see this added.

### LINKED ISSUE #125: Map misspellings → corrections, not just a single-term dictionary
## What's the idea?

Right now the dictionary / custom vocabulary is a list of correct terms (e.g., `Claude Code`). The LLM is supposed to fix "close misspellings" of those terms during post-processing (per issue #49's system prompt: *"When the transcript already contains a close misspelling of a name or term from context or vocabulary, correct it"*).

In practice, this does not always work. I added `Claude Code` to my dictionary, and the transcript still came through with the wrong spelling — the model did not treat what I said as close enough to the dictionary entry to correct it.

## Proposal

Let dictionary entries be an explicit mapping from misheard → correct, not just a single term:

| Heard | Corrected |
|---|---|
| `cloud code`, `clod code`, `clawed code` | `Claude Code` |
| `next message` | `nextMessage` |
| `DDX web` | `DDXWeb` |

The LLM still does the final judgment call, but now it has concrete pairs to pattern-match against. A transcript of "cloud code" becomes a much stronger signal to emit "Claude Code" than asking the model to infer it from a one-sided vocabulary list.

## Prior art: Wispr Flow

This is exactly how Wispr Flow exposes it — a dedicated "Edit replacement" UI with a heard-form field on the left and the correct form on the right. Simple, explicit, and scales to any number of entries.

Wispr Flow Edit replacement dialog showing 'Power Reels parties.' mapping to 'Power Wheels Parties'

Note the example in the bottom-left (`btw → by the way`) — it doubles as text expansion, which is a useful side benefit but not the main ask here.

## Why a single-term dictionary falls short

- Close-misspelling detection is implicit — it depends on the model deciding the transcript is "close enough." With unusual brand names, compound terms, or homophones, it often is not.
- No way for the user to teach the app about specific failure modes they keep hitting.
- No way to encode intent — `Claude` vs `clod` is not a spelling variant, it is a mis-hearing. A mapping captures that directly.

## How it could surface in the UI

- Dictionary entry form gets a second optional field: "Also correct these heard-as variants" (comma-separated or one-per-line).
- Entries without variants behave exactly as today (backward compatible).
- Under the hood, variants get appended to the prompt as pairs the model is asked to normalize.

## Alternatives considered

- Editing the system prompt manually (per issue #49) — works, but does not scale and is invisible to anyone else using the app.
- Post-dictation find-and-replace — lossy, and would clobber legitimate uses of the heard-as form.

## Related

- #49 — Improving the default system prompt (shows the current "close misspelling" mechanism)
- #120 — Wispr-Flow-style self-correction cleanup (different concept; orthogonal to this)

---

**Environment**
- FreeFlow: 0.3.0 — `build-20260421-154112-1167337`
- macOS: 26.3.1 (arm64, Apple M2 Ultra)

---

*This issue was dictated using [FreeFlow](https://github.com/zachlatta/freeflow) and written up with [Claude Code](https://claude.com/claude-code).*

### LINKED ISSUE #196: Feature Request: "Prompt Mode" — summarize/condense dictated text before pasting
What’s the feature?

A toggleable mode (or secondary shortcut) where instead of pasting cleaned-up dictation verbatim, FreeFlow passes the transcript through a summarization/compression step first — outputting a tight, intent-preserving 
version of what was said.

Why it matters

Wispr Flow has a version of this. When you’re dictating directly into an AI tool (Claude, ChatGPT, Cursor, etc.), you naturally speak in a rambling, conversational way — filler, repetition, false starts. The current cleanup does a great job fixing grammar and filler words, but the output is still the full length of what you said. For AI prompts specifically, a concise prompt performs better than a verbose one.

Proposed behavior

	•	Normal mode: cleanup as-is (current behavior, untouched)
	•	Prompt mode: cleanup → condense → paste a tight 1–3 sentence version that preserves full intent
	•	Could be a settings toggle, a separate hotkey combo, or context-triggered (e.g., when active app is a browser pointing to claude.ai, chatgpt.com, etc.)
Workaround I’m currently using
Custom system prompt tweak — works partially but applies globally and doesn’t intelligently compress, just cleans.

### LINKED ISSUE #200: Spacing, capitalization, and punctuation are context-blind — no awareness of text surrounding the cursor

### LINKED ISSUE #237: Edit Mode silently falls back to dictation in Electron (VSCode) & Chromium (Gmail) — overwrites selection instead of transforming
## Summary
Edit Mode works in native Accessibility (AX) apps like TextEdit, but in Electron/Chromium apps it just transcribes the spoken instruction and **overwrites the selected text with the literal transcript** instead of transforming it.

## Environment
- FreeFlow **1.1.0** (stable) — reproduced
- FreeFlow **Dev** build — same behavior
- macOS Sonoma 14.3.1
- Edit Mode: **Auto**
- Accessibility permission re-granted + app restarted

## Steps to reproduce
1. Select text in VSCode (or Gmail in a browser)
2. Trigger the normal dictation shortcut
3. Say e.g. "make this shorter"

**Expected:** the selection is transformed per the instruction.
**Actual:** the instruction is transcribed and replaces the selection verbatim.

## Works / doesn't
- TextEdit (native AX): works
- VSCode (Electron): fails
- Gmail web (Chromium): fails
- Plain dictation works everywhere.

## Notes
Consistent with the known AX limitation — Electron/Chromium don't expose the current selection via the Accessibility API, so the Edit Mode intent resolver falls through to dictation. The clipboard-fallback work in #104 that would have addressed this was closed/abandoned, and there's no shipped toggle or fix in stable or the Dev build. Could a clipboard-read fallback (or a manual Edit-Mode trigger that forces Cmd+C capture) be reconsidered for these apps?

### LINKED ISSUE #253: URLSession timeoutIntervalForResource=30 in LLMAPITransport silently caps user timeout overrides
## Problem
`LLMAPITransport` pins `timeoutIntervalForResource = 30` on its `URLSessionConfiguration`, which is a session-level hard cap with no per-request override, so the README-documented `transcription_timeout_seconds` / `post_processing_timeout_seconds` / `context_request_timeout_seconds` defaults are silently ignored for any transfer >30s.

## Read cite
`Sources/LLMAPITransport.swift:13` — `configuration.timeoutIntervalForResource = 30`

## Steps
1. Configure a slow local model per README "Configure longer timeouts for local models":
   ```bash
   defaults write com.zachlatta.freeflow transcription_timeout_seconds -float 120
   ```
2. Point transcription at a slow local Whisper endpoint whose upload+transcribe exceeds 30s.
3. Dictate a long clip.

## Expected
Request allowed to run up to 120s per README + `TranscriptionService.transcriptionTimeoutSeconds` (`TranscriptionService.swift:12-15`).

## Actual
URLSession aborts the resource at 30s regardless of the user override; per-request `request.timeoutInterval = 120` does not override `timeoutIntervalForResource` (Apple: this property has no per-request equivalent).

## Environment
FreeFlow 1.1.0 (CHANGELOG), macOS 13.0+ target (Makefile `-target arm64-apple-macosx13.0`).

Fix: derive the session's `timeoutIntervalForResource` from the same UserDefaults keys, or expose a per-call session factory.

Thanks for maintaining zachlatta/freeflow!

swimmer-303:
This PR bundles a set of fixes and features developed while dogfooding FreeFlow daily. Each change is a separate commit with a descriptive message, so it can be reviewed commit-by-commit — and I'm happy to split any subset into standalone PRs if that's easier to review or land.

## Fixes

**Honor configured timeouts instead of capping transfers at 30s** (fixes #253)
`LLMAPITransport` pinned `timeoutIntervalForResource = 30` at the session level, silently overriding the README-documented `transcription_timeout_seconds` / `post_processing_timeout_seconds` / `context_request_timeout_seconds` overrides — any transfer over 30s was killed regardless of settings, breaking slow local models. Sessions now derive their whole-transfer budget from each request's configured timeout (30s floor), cached per timeout value so TLS connection reuse is preserved. Uploads keep their fresh-session-per-call poisoning defense.

**Edit Mode no longer silently overwrites selections in Electron/Chromium apps** (fixes #237)
Chromium-based apps build their accessibility tree lazily, so `AXSelectedText` reads failed and Edit Mode fell back to dictation — overwriting the user's selection in VS Code, Gmail-in-Chrome, etc. FreeFlow now sets Electron's `AXManualAccessibility` opt-in on the frontmost app before reading the selection (harmless no-op elsewhere), plus Chromium's `AXEnhancedUserInterface` for known browsers only (scoped narrowly since that attribute can interact badly with window-manager utilities). The first read after activation gets one 100 ms retry because the tree builds asynchronously.

## Features

**Mishearing → correction mappings in custom vocabulary** (closes #125, `help wanted`)
Vocabulary entries can map heard forms to the intended term with an arrow — `cloud code -> Claude Code` — and multiple heard forms can share one correction: `cloud code | clod code -> Claude Code`. Pairs are passed to the post-processing model as an explicit corrections block (with a guard against over-applying); plain entries behave exactly as before. Parser verified against plain entries, single/multi-variant mappings, mixed comma-separated lines, `=>`, and malformed input.

**Cursor-aware spacing, capitalization, and punctuation** (fixes #200)
Dictating into the middle of a sentence used to produce a stray leading capital, a trailing period, and text jammed against existing words. FreeFlow now reads up to 200 chars before / 80 after the insertion point (via `AXStringForRange` with an `AXValue`-slice fallback; secure fields are never read) and uses it three ways: the cleanup prompt gets the surrounding text with rules to continue the sentence flow; a separating space is prepended deterministically when pasting directly after a word character or closing punctuation (skipped in Edit Mode); and the auto trailing space after sentence punctuation is suppressed when the following text already provides separation.

**Prompt Mode** (closes #196)
Off (default) / Always / Only in AI apps. Condenses rambling dictation into a tight, intent-preserving prompt before pasting. Auto mode detects Claude, ChatGPT, Cursor, and similar tools from the frontmost app's bundle ID and window title. Condensation is a section appended to the cleanup system prompt in the same LLM pass, so it adds no latency; the instruction-execution guard still applies, and Edit Mode is unaffected.

**Configurable feedback sounds** (closes #107)
The hardcoded Tink/Pop/Basso start/stop/error sounds are now per-event settings with pickers and preview buttons covering all 14 built-in macOS alert sounds. Defaults match the old behavior.

## Performance

**Prewarm API connections while recording**
Transcription uploads deliberately use a fresh URLSession per upload, which put a full DNS + TCP + TLS handshake (~100–300 ms) on the critical stop-to-paste path of every dictation. A one-shot session is now created and its connection opened when recording starts; the next upload to the same host consumes it — still exactly one session per upload, so the poisoning defense is unchanged. The shared data session is warmed the same way for post-processing/context requests.

**Start transcription before capture teardown**
`AVCaptureSession.stopRunning()` can take 100–300 ms and ran before the stop completion fired. The audio cut-off point (sample-buffer delegate removal) and the queued-tail drain are byte-for-byte unchanged; only the slow teardown now happens after the completion, so the upload starts immediately.

**Latency instrumentation**
Transcription uploads and LLM requests now log durations to the existing `com.zachlatta.freeflow` os_log subsystem, so pipeline latency is visible in Console.

## UI

**Menu bar panel redesign + live waveform icon**
The menu bar extra now uses a window-style panel, and while recording the menu bar icon renders a small live waveform driven by the input level.

## Notes

- The near-cursor recording overlay (#212) is intentionally **not** included — #242 by @mvanhorn already covers it and merges cleanly alongside this branch.
- Verified with `make` builds throughout; pure logic (vocabulary parser, smart-spacing) covered by standalone test scripts; running as my daily driver since building it.

## Summary by CodeRabbit

* **New Features**
  * Added Prompt Mode to condense dictation before pasting, plus cursor-aware cleanup for smoother text insertion.
  * Expanded Custom Vocabulary to support mishearing corrections and multiple heard variants.
  * Added configurable start, stop, and error sounds in Settings.
  * Updated the menu bar indicator to show live recording activity more clearly.

* **Bug Fixes**
  * Improved dictation in some apps by better handling text selection and accessibility.
  * Reduced stop-to-paste delay and removed an occasional timeout issue that could interrupt longer dictations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ This project uses semantic versioning for public releases. Use `MAJOR.MINOR.PATC
 
 ## [Unreleased]
 
+### Improved
+
+- Faster stop-to-paste latency: API connections to the transcription and post-processing hosts are now opened while you are still speaking, so the upload no longer pays DNS + TLS handshake time after you release the shortcut.
+- Faster stop-to-paste latency: the transcription upload now starts immediately when recording stops, instead of waiting for the audio capture session to finish tearing down.
+- Transcription uploads and post-processing requests now log their durations, making pipeline latency visible in Console.
+
 ### Added
 
 - Custom vocabulary entries can now map specific mishearings to the intended term with an arrow (`cloud code -> Claude Code`), and multiple heard forms can share one correction (`cloud code | clod code -> Claude Code`). Plain entries behave exactly as before.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,17 @@ This project uses semantic versioning for public releases. Use `MAJOR.MINOR.PATC
 - `MINOR` changes add user-visible features and improvements.
 - `PATCH` changes fix bugs, polish existing behavior, or make small internal improvements.
 
+## [Unreleased]
+
+### Added
+
+- Custom vocabulary entries can now map specific mishearings to the intended term with an arrow (`cloud code -> Claude Code`), and multiple heard forms can share one correction (`cloud code | clod code -> Claude Code`). Plain entries behave exactly as before.
+- The recording start, stop, and error feedback sounds are now configurable in Settings, with a picker and preview button for each event across the full set of built-in macOS alert sounds.
+
+### Fixed
+
+- Fixed the configured transcription, post-processing, and context timeouts being silently capped at 30 seconds by a session-level resource timeout, which broke long transfers to slow local models.
+
 ## [1.1.0] - 2026-06-03
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project uses semantic versioning for public releases. Use `MAJOR.MINOR.PATC
 
 ### Added
 
+- Dictation is now aware of the text surrounding the cursor: cleanup matches the sentence flow when inserting mid-sentence (no stray leading capitals or trailing periods), a separating space is added automatically when inserting directly after a word, and the trailing space after sentence punctuation is skipped when the following text already provides one.
 - Custom vocabulary entries can now map specific mishearings to the intended term with an arrow (`cloud code -> Claude Code`), and multiple heard forms can share one correction (`cloud code | clod code -> Claude Code`). Plain entries behave exactly as before.
 - The recording start, stop, and error feedback sounds are now configurable in Settings, with a picker and preview button for each event across the full set of built-in macOS alert sounds.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project uses semantic versioning for public releases. Use `MAJOR.MINOR.PATC
 
 ### Added
 
+- Prompt Mode: optionally condense rambling dictation into a tight, intent-preserving prompt before pasting — set to Always, or Only in AI apps (auto-detects Claude, ChatGPT, Cursor, and similar tools from the active app and window title). Runs in the same cleanup pass, so it adds no latency.
 - Dictation is now aware of the text surrounding the cursor: cleanup matches the sentence flow when inserting mid-sentence (no stray leading capitals or trailing periods), a separating space is added automatically when inserting directly after a word, and the trailing space after sentence punctuation is skipped when the following text already provides one.
 - Custom vocabulary entries can now map specific mishearings to the intended term with an arrow (`cloud code -> Claude Code`), and multiple heard forms can share one correction (`cloud code | clod code -> Claude Code`). Plain entries behave exactly as before.
 - The recording start, stop, and error feedback sounds are now configurable in Settings, with a picker and preview button for each event across the full set of built-in macOS alert sounds.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This project uses semantic versioning for public releases. Use `MAJOR.MINOR.PATC
 
 ### Fixed
 
+- Fixed Edit Mode silently falling back to dictation (overwriting the selection) in VS Code, Gmail-in-Chrome, and other Electron/Chromium apps. Those apps build their accessibility tree lazily, so FreeFlow now asks them to activate it before reading the selection.
 - Fixed the configured transcription, post-processing, and context timeouts being silently capped at 30 seconds by a session-level resource timeout, which broke long transfers to slow local models.
 
 ## [1.1.0] - 2026-06-03
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,25 @@
+## Agent Guardrails
+
+You are running as an agent. Do not run destructive commands without explicit permission.
+
+### Git
+- Never run `git push --force`, `git push -f`, or rewrite shared history
+- Never delete branches (local or remote) without explicit permission
+- Never commit `.env*` files, API keys, tokens, or credentials
+
+### Filesystem
+- Never run `rm -rf` on anything outside the current project directory
+- Never modify files outside the current working directory without explicit permission
+- Never edit `.env`, `.env.local`, or any `.env.*` file without explicit permission
+
+### Dependencies
+- Ask before installing new packages or running `npm install <pkg>`
+- Ask before upgrading major versions or modifying lockfiles by hand
+
+### Database
+- Never run database migrations, drops, or schema changes without explicit permission
+
+### Behavior
+- When uncertain, stop and ask rather than guess
+- Summarize planned changes before applying anything non-trivial
+- Don't disable, skip, or comment out tests, linting, or type checks to make things pass
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ FreeFlow is a free Mac dictation app inspired by [Wispr Flow](https://wisprflow.
 - **Custom shortcuts:** Customize both hold-to-talk and toggle dictation shortcuts. If your toggle shortcut extends your hold shortcut, you can start in hold mode and press the extra modifier keys to latch into tap mode without stopping the recording.
 - **Context-aware cleanup:** FreeFlow can read nearby app context so names, terms, and phrases are spelled correctly when you dictate into email, terminals, docs, and other apps. It also reads the text around your cursor, so dictating into the middle of a sentence continues the sentence naturally with correct capitalization, punctuation, and spacing.
 - **Custom vocabulary:** Add names, jargon, and project-specific words that FreeFlow should preserve during cleanup. Entries can also map specific mishearings to the intended term with an arrow — `cloud code -> Claude Code` — and multiple heard forms can share one correction: `cloud code | clod code -> Claude Code`.
+- **Prompt Mode:** Optionally condense rambling dictation into a tight prompt that preserves your full intent — always, or automatically when dictating into AI tools like Claude, ChatGPT, and Cursor.
 - **OpenAI-compatible providers:** Use Groq by default, or configure a custom model and API URL in settings.
 
 ## Edit Mode
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ FreeFlow is a free Mac dictation app inspired by [Wispr Flow](https://wisprflow.
 
 - **Custom shortcuts:** Customize both hold-to-talk and toggle dictation shortcuts. If your toggle shortcut extends your hold shortcut, you can start in hold mode and press the extra modifier keys to latch into tap mode without stopping the recording.
 - **Context-aware cleanup:** FreeFlow can read nearby app context so names, terms, and phrases are spelled correctly when you dictate into email, terminals, docs, and other apps.
-- **Custom vocabulary:** Add names, jargon, and project-specific words that FreeFlow should preserve during cleanup.
+- **Custom vocabulary:** Add names, jargon, and project-specific words that FreeFlow should preserve during cleanup. Entries can also map specific mishearings to the intended term with an arrow — `cloud code -> Claude Code` — and multiple heard forms can share one correction: `cloud code | clod code -> Claude Code`.
 - **OpenAI-compatible providers:** Use Groq by default, or configure a custom model and API URL in settings.
 
 ## Edit Mode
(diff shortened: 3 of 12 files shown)


swimmer-303:
For what it's worth, I've been daily driving this build as my main dictation setup since making these changes — dictating into Claude Code, email, and browser text fields all day. The latency improvements are noticeable, and Prompt Mode + the vocabulary mappings have held up well in real use. Happy to split this into smaller PRs if that's easier to review.

marcbodea:
Thanks for offering to split this up. Reviewing commit-by-commit helps, but independent PRs would give us clean review, merge, and revert boundaries.

Could you reduce this PR to the timeout fix for #253, then open separate PRs for:
1. Custom vocabulary correction mappings (#125)
2. Configurable feedback sounds (#107)
3. Electron/Chromium Edit Mode accessibility fix (#237)
4. Cursor-aware spacing, capitalization, and punctuation (#200)
5. Prompt Mode (#196)
6. API connection prewarming and latency instrumentation, based on the timeout fix
7. Starting transcription before capture-session teardown
8. Menu-bar panel redesign/live waveform
9. Settings-sidebar redesign

CLAUDE.md should also be a separate documentation-only PR, or omitted from this series. Please move each README/changelog entry into the PR that introduces that behavior.

The main concern is the Add configurable feedback sounds, menu bar panel redesign, changelog commit: it currently combines sounds, the menu-bar overhaul, the settings-sidebar redesign, vocabulary documentation, changelog changes, and agent guardrails. Those should not share one review or merge boundary.

The existing review findings can then be addressed in their corresponding PRs. This should let us land the small bug fixes quickly while reviewing the larger UI and performance changes independently.

swimmer-303:
Thanks @marcbodea — that makes sense. I've split this into independent PRs, each with clean review/merge/revert boundaries, and moved each README/changelog entry into the PR that introduces its behavior. Closing this in favor of them:

**Bug fix**
- #263 — Honor configured timeouts instead of capping transfers at 30s (fixes #253)

**Performance** (based on the timeout fix)
- #264 — Prewarm API connections while recording + latency instrumentation (**stacked on #263**; the diff reduces to just the prewarming/instrumentation once #263 merges)
- #266 — Start transcription before capture-session teardown

**Features**
- #269 — Custom vocabulary mishearing→correction mappings (closes #125)
- #270 — Configurable feedback sounds (closes #107)
- #268 — Prompt Mode (closes #196)
- #267 — Cursor-aware spacing, capitalization, and punctuation (fixes #200)

**Edit Mode fix**
- #265 — Activate lazy Chromium/Electron accessibility trees for Edit Mode (fixes #237)

**UI**
- #271 — Menu bar panel redesign + live waveform icon
- #272 — Settings sidebar redesign

**Docs**
- #273 — CLAUDE.md agent guardrails (documentation-only)

The combined "sounds + menu-bar + sidebar" commit that was the main concern is now three separate PRs (#270, #271, #272) with no shared review or merge boundary. Each PR builds independently on top of `main`. Existing review findings can be addressed in the corresponding PR.

Issue #314, opened 2026-09-02
First toggle-stop press is swallowed after a menu-bar start


JaPossert:
**Everything below this line is AI generated.**

Context: This is part of a wave of issues - #308 to #322, written up as I am concluding extending it for my personal purposes for now. I had also opened PR #250, which you merged some time ago.

Working demo rather than a pull request: the code runs and is used daily, and it is readable at [`PLUGINS/`](https://github.com/JaPossert/freeflow/tree/demo/plugins-vibe-demo/PLUGINS) on my fork, with the full integration on [`feat/async-target-paste`](https://github.com/JaPossert/freeflow/tree/feat/async-target-paste). Take, adapt or ignore any of it.

---

### Problem

Start a recording from the menu bar button in toggle mode, then press the toggle shortcut to stop it. The first press does nothing.

`DictationShortcutSessionController.beginManual(mode:)` sets `toggleStopArmed = false`, but the toggle branch of `handle(event:isTranscribing:)` requires `toggleStopArmed` to be true before it will return `.stop`. A session begun without a physical press and release cycle therefore never arms, and the first shortcut press is swallowed silently.

### Fix

```swift
func beginManual(mode: RecordingTriggerMode) {
    activeMode = mode
    // Started without a physical key press/release cycle, e.g. the menu bar
    // button, so arm stop immediately.
    toggleStopArmed = (mode == .toggle)
}
```

One line. On https://github.com/JaPossert/freeflow/tree/feat/async-target-paste, and easily cherry-picked on its own.

Issue #308, opened 2026-09-02
Caret-exact insertion is impossible inside terminals (unresolved, with findings)


JaPossert:
> **Human notes**
>
Make not something to solve but to be reminded of.
---

**Everything below this line is AI generated.**

This only makes sense alongside #309, which is the general case: put the transcript back where the caret *was*, not where the caret is by the time the text is ready. Everything here is that same requirement meeting a surface that has no caret to record.

Context: This is part of a wave of issues - #308 to #322, written up as I am concluding extending it for my personal purposes for now. I had also opened PR #250, which you merged some time ago.

I have tried to keep your PRs overviewable and thus only made a link to my vibecoded implemtation that I use daily

Working demo rather than a pull request: the code runs and is used daily, and it is readable at [`PLUGINS/`](https://github.com/JaPossert/freeflow/tree/demo/plugins-vibe-demo/PLUGINS) on my fork, with the full integration on [`feat/async-target-paste`](https://github.com/JaPossert/freeflow/tree/feat/async-target-paste). Take, adapt or ignore any of it.

---

### What I hoped for

Dictation should put the text back where the caret was when I released the key, not merely in the right window. I often keep working in the same app at a different spot while the transcription runs.

For Accessibility-capable text fields this works: pin the focused `AXUIElement` **and** its `kAXSelectedTextRange` at stop time, then restore the range before writing `kAXSelectedTextAttribute`. Implemented in `Sources/TextDelivery.swift` on https://github.com/JaPossert/freeflow/tree/feat/async-target-paste.

### Where it does not work, and why

**Terminals.** Every delivery into Terminal.app in my log resolves to synthesised keystrokes, never the Accessibility path. The reason is structural rather than a bug: a terminal emulator draws its content, it does not expose an editable Accessibility element with a selection. `kAXSelectedTextRange` is not settable there, and the shell or TUI owns the cursor entirely. There is no coordinate for "that exact spot" to restore, so the text necessarily lands wherever the cursor happens to be at delivery time.

I looked for a way around it and did not find one:

- `ioctl(fd, TIOCSTI)` would inject straight into the tty, which would be exact. macOS returns `EPERM` for a non-root process; verified with a pty round-trip probe.
- Terminal.app's AppleScript `do script` executes rather than inserting, so it cannot place text mid-prompt.
- `CGEvent.postToPid` with unicode does deliver, but only at the current cursor.

### Status

[...]

PR #315, opened 2026-09-02
Deliver transcripts back to the element they were dictated into, and let dictations overlap

JaPossert:
> **Human notes**
>
> 

&nbsp;

---

**Everything below this line is AI generated.**

I also opened #250, which you merged, and #252, which is still open. This pull request implements the work described in issues #308 to #314.

---

## What this does

Dictation no longer makes you stand still while it transcribes.

**Delivery goes back to where the words came from.** At stop time the fork pins the focused `AXUIElement` and its `kAXSelectedTextRange`, then walks an ordered ladder to put the transcript there: an Accessibility write into that exact element, a Cmd-V addressed to the process, unicode key events addressed to the process, and finally the clipboard with a panel saying so. Only the opt-in rung activates anything, so switching window, app or Desktop mid-transcription is safe. Today's `activate` plus Cmd-V survives as that opt-in rung.

**Dictations can overlap.** `isTranscribing` was a single boolean and `transcriptionTask` a single task, so the next dictation had to wait. Sessions are now keyed by UUID, each carrying its own delivery target, with `isTranscribing` derived from the live set. Against a local model that removes a 10 to 25 second wait per 30 seconds of speech.

**The overlay tells the truth about what is still running.** It stays up until the last transcription has delivered, with a count badge covering every transcription in flight plus the recording in progress, shown from two onwards.

**The recording indicator shows loudness.** A scrolling meter rather than a waveform whose motion is independent of the level.

## Things found along the way that may matter to you regardless of this PR

- An Accessibility write into a Chromium or Electron host returns `.success` while the text never reaches the editable surface. Very likely the same root cause as #237. The insert now verifies by measuring the element before and after.
- Chromium reads a synthesised key event by its virtual key code and ignores the attached unicode, so injected text arrives as a stray keypress.
- Chromium only handles key events while it is the active application, so there is no way to deliver into a backgrounded Electron app without stealing focus. The fork stops pretending otherwise and falls through to the clipboard.
- Three hazards that only exist once dictations can overlap: the shared `AudioRecorder` being cleaned up by an earlier session's completion, the clipboard snapshot capturing a previous transcript, and `endCriticalDictationActivity()` firing early.

## Honest limits

- Caret-exact insertion cannot work in terminals: no editable Accessibility element, so no range to restore. Written up in #308 with what I tried, including `TIOCSTI`, which macOS refuses for non-root.
- Backgrounded Chromium hosts fall through to the clipboard, by design rather than by accident.

## How it was built and tested

Written with AI assistance, then used daily against a local Qwen3 ASR server. Verified in Claude Code, pi, Brave, Terminal, Tana and native text fields, across windows, applications and Spaces. Every delivery writes its whole ladder to a `delivery.log` beside the run log, which is where most of the findings above came from.

Happy to split this into smaller pieces if that suits review better. #314 in particular is a one-line fix that stands entirely on its own.

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous transcript delivery to the original target using accessibility insertion, keystrokes, paste, or clipboard fallback.
  * Added support for overlapping dictation sessions and transcription prefetching.
  * Added loudness meter, transcription progress, and clipboard-waiting panels.
  * Added settings for delivery behavior, background transcription, overlapping dictation, focus handling, and transcription timeouts.

* **Bug Fixes**
  * Fixed missed first toggle-stop presses and improved delivery failure detection.
  * Bare Escape now passes through unless Fn is held.

* **Performance**
  * Increased transcription request time limits for longer recordings.

### DIFF
--- a/DEVLOG/async-delivery.md
+++ b/DEVLOG/async-delivery.md
@@ -123,6 +123,39 @@ which is why short test recordings produce 0-byte
 `prefetch_transcript_*.txt` files. Not waiting for any of it is the entire
 point of this fork.
 
+## Field results, 2026-09-02
+
+Works: Claude Code, pi, Brave, and across windows and apps, including
+continuing to type elsewhere while it delivers. Does not work: Tana, Paseo,
+and a bare shell prompt in Terminal, where the text simply disappears.
+
+Three changes came out of that session:
+
+- **Exact caret, not just the window.** `DeliveryTarget` now also pins
+  `kAXSelectedTextRange`, and the insert restores that range before writing.
+  Continuing to work at a different spot in the same document no longer sends
+  the transcript to wherever the caret happens to be.
+- **Verified writes.** An AX write into Chromium or Electron can return
+  `.success` while the text never reaches the editable surface, which is
+  precisely how a transcript vanishes in Tana. The insert now reads
+  `kAXNumberOfCharacters`, falling back to `kAXValue`, before and after, and
+  treats "no growth" as failure so the ladder falls through to the keystroke
+  tier. Elements exposing neither attribute are accepted unverified.
+- **The overlay outlives one session.** `dismissOverlayIfIdle()` only takes the
+  overlay down when no transcriptions remain, and
+  `pendingTranscriptionCount` drives a count badge to the left of the
+  processing bars, so a queue of several is legible.
+
+## Signing
+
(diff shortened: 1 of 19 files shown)


JaPossert:
Closing this: I do not want to hand over a vibe-coded pull request. The work stays available as a branch to read and cherry-pick from, and the reasoning lives in issues #308 to #314 instead. Sorry for the noise.

